PR #5, opened 2025-02-05
chore: Remove manifest script to manage via `iota move manage-package`

DaughterOfMars:
## Description

Removes the janky ts script that rewrites the manifests in favor on the newer method using managed packages.


packages/denylist/Move.lock
```diff
@@ -1,22 +1,17 @@
 # @generated by Move, please check-in and do not edit manually.
 
 [move]
-version = 1
-manifest_digest = "1C84167604CF87169A914020CB310C1A1B60DE762E11FE6EE258940655B2498B"
+version = 2
```
Thoralf-M, line 4: Just wondering, should we commit these changes?

packages/iota-names/sources/name_record.move
```diff
@@ -2,11 +2,7 @@
 // Modifications Copyright (c) 2025 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-/// The `NameRecord` is a struct that represents a single record in the
-/// registry.
-/// Can be replaced by any other data structure due to the way `NameRecord`s are
-/// stored and managed. IotaNames has no direct and permanent dependency on this
-/// module.
```
Thoralf-M, line 9: Why remove this part? Sounds interesting to me

scripts/utils/utils.ts
```diff
@@ -51,6 +59,31 @@ export const publishPackage = (txb: Transaction, path: string, configPath?: stri
...
+		...(configPath ? ['--client.config', configPath] : []),
+		'manage-package',
+		'--environment',
+		active_env,
+		'--network-id',
+		chain_id,
+		'--original-id',
+		package_id,
+		'--latest-id',
+		package_id,
```
Thoralf-M, line 86: Is this function only called the very first time? Otherwise the latest-id and the original-id need to be different?

packages/denylist/Move.lock
DaughterOfMars: It will get rewritten when we use the correct stuff anyway so I don't see any real problem

packages/iota-names/sources/name_record.move
DaughterOfMars: Because it's wrong haha

scripts/utils/utils.ts
DaughterOfMars: Yes, they should be different after the first publish. I don't know the best way to do that yet so I left it as-is.

Thoralf-M: Put a TODO at least

packages/iota-names/sources/name_record.move
Thoralf-M: Depends if one sees `IotaNames` as package or just the struct

Thoralf-M: But seems to be the only file with such a comment, fine to have it removed, one can just check the imports in the iota_names package to see the same

scripts/utils/utils.ts
DaughterOfMars: done

Thoralf-M (approved):

scripts/utils/utils.ts
Thoralf-M: As reference for others https://github.com/iotaledger/iota-names/issues/17

packages/denylist/Move.toml
```diff
@@ -2,18 +2,10 @@
...
 edition = "2024.beta"
-# mainnet
-published-at = "0xc967b7862d926720761ee15fbd0254a975afa928712abcaa4f7c17bb2b38d38b"
-
-# testnet
-# published-at="a86c05fbc6371788eb31260dc5085f4bfeab8b95c95d9092c9eb86e63fae3d49"
 
 [dependencies]
-Iota = { git = "https://github.com/iotaledger/iota.git", subdir = "crates/iota-framework/packages/iota-framework", rev = "framework/mainnet", override = true }
+Iota = { git = "https://github.com/iotaledger/iota.git", subdir = "crates/iota-framework/packages/iota-framework", rev = "64e8125", override = true }
```
thibault-martinez: why not testnet?

packages/denylist/Move.toml
```diff
@@ -2,18 +2,10 @@
...
 IotaNames = { local = "../iota-names" }
 
 [addresses]
-denylist = "0xc967b7862d926720761ee15fbd0254a975afa928712abcaa4f7c17bb2b38d38b"
-
-# testnet
-# denylist = "a86c05fbc6371788eb31260dc5085f4bfeab8b95c95d9092c9eb86e63fae3d49"
+denylist = "0x0"
```
thibault-martinez, line 11: 0x0?

packages/iota-names/sources/controller.move
```diff
@@ -81,7 +81,7 @@ entry fun set_user_data(
     let domain = nft.domain();
 
     registry.assert_nft_is_authorized(nft, clock);
-    let key_bytes = *key.bytes();
+    let key_bytes = *key.as_bytes();
```
thibault-martinez: Would be best to do these separate tbh

packages/denylist/Move.toml
DaughterOfMars: I could do that. Wasn't sure what branch we want to use tbh

packages/denylist/Move.toml
DaughterOfMars: Yes, these should all use 0x0 and be managed via their lock files

packages/iota-names/sources/controller.move
DaughterOfMars: Sure but it's not a big change

packages/denylist/Move.toml
DaughterOfMars: it's done btw

thibault-martinez (approved):

PR #8, opened 2025-02-07
Give admin mutable access to config object

Alex6323:
not sure yet, up for discussion, but could make sense


Alex6323:
not needed anymore

PR #9, opened 2025-02-07
disallow invalid prices on config construction

Alex6323:
Currently it's possible for the admin to create a config with invalid (out-of-range) prices. This PR fixes that.


Alex6323:
not needed anymore

PR #15, opened 2025-02-13
fix(CI): Fix basic CI

DaughterOfMars:
## Description

Fixes the basic CI jobs by using cargo install with the git repo rather than homebrew which is not available (yet).


Alex6323 (approved):

DaughterOfMars:
I changed the base to the other PR as this should be merged afterward, and regenerated the lock file with updated packages

PR #19, opened 2025-02-17
feat: apply dprint fmt and add CI workflow

thibault-martinez:


.github/workflows/dprint.yaml
```diff
@@ -14,10 +14,10 @@ concurrency:
 jobs:
   run:
     name: Check dprint format
-    runs-on: ubuntu-latest
```
Alex6323: why? isn't that relatively cheap?

thibault-martinez: To avoid having to install it every time

.github/workflows/iota-names-combine-execute.yaml
```diff
@@ -5,42 +5,42 @@ on:
   workflow_dispatch:
     inputs:
       rpc:
-        description: 'RPC for iota client'
+        description: "RPC for iota client"
```
Alex6323:
```suggestion
        description: "RPC for the IOTA client"
```

Alex6323 (approved):

Thoralf-M (approved):

PR #23, opened 2025-02-17
fix(nodejs): fix package versions, imports, lint:fix

Thoralf-M:
Tested by running

```
export NETWORK=localnet
pnpm ts-node init/init.ts
```


scripts/transactions/addresses.json
Thoralf-M, line 1: I don't see this referenced anywhere in the code, I guess we should just delete it

sdk/src/iota-names-client.ts
```diff
@@ -99,15 +99,15 @@ export class IotaNamesClient {
 	}
 
 	async getNameRecord(name: string): Promise<NameRecord | null> {
-		if (!isValidIotaName(name)) throw new Error('Invalid IOTA name');
+		if (!isValidIotaAddress(name)) throw new Error('Invalid IOTA name');
```
DaughterOfMars: Wait, this can't be right

scripts/transactions/addresses.json
thibault-martinez, line 1: What is this used for do you know?

Thoralf-M: No, in the current codebase I don't see any reference to this anywhere

sdk/src/iota-names-client.ts
Thoralf-M: Good catch, they had `import { isValidSuiNSName, normalizeSuiNSName } from '@mysten/sui/utils';` but we don't have this in the iota-sdk anymore, I guess we should just add its definition here for now

Thoralf-M: https://github.com/MystenLabs/ts-sdks/blob/12547c3bd4c6b2970e4b569cdba9d0da3d5f37d2/packages/typescript/src/utils/suins.ts#L9

DaughterOfMars: Yeah definitely, it's weird that it was in that package to begin with

sdk/src/utils.ts
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Mysten Labs, Inc.
+// Modifications Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+const IOTA_NAME_REGEX =
+	/^(?!.*(^(?!@)|[-.@])($|[-.@]))(?:[a-z0-9-]{0,63}(?:\.[a-z0-9-]{0,63})*)?@[a-z0-9-]{0,63}$/i;
+const IOTA_NAMES_DOMAIN_REGEX = /^(?!.*(^|[-.])($|[-.]))(?:[a-z0-9-]{0,63}\.)+iota$/i;
+const MAX_IOTA_NAME_LENGTH = 235;
```
DaughterOfMars, line 8: what to heck???? there's no way this regex is necessary 😆 obviously no need to fix it here but we should remember to look at this later

DaughterOfMars (approved):

sdk/src/iota-names-transaction.ts
```diff
@@ -4,12 +4,13 @@
...
 import { bcs } from '@iota/iota-sdk/bcs';
 import type { Transaction } from '@iota/iota-sdk/transactions';
-import { isValidIotaName, normalizeIotaName, IOTA_CLOCK_OBJECT_ID } from '@iota/iota-sdk/utils';
+import { IOTA_CLOCK_OBJECT_ID } from '@iota/iota-sdk/utils';
 
 import { ALLOWED_METADATA } from './constants.js';
 import { isNestedSubName, isSubName, validateYears } from './helpers.js';
 import type { IotaNamesClient } from './iota-names-client.js';
 import type { ObjectArgument } from './types.js';
+import { isValidIotaName, normalizeIotaName } from './utils.js';
```
thibault-martinez, line 13: Is this import correct?

Thoralf-M: I had to add it, since they originally used `@mysten/sui/utils` and in `@iota/iota-sdk/utils` we don't have such functions

thibault-martinez: Ah, I was looking for `utils.js` but it's after being compiled then, my bad

thibault-martinez: Where they removed or moved somewhere else? Should we add them to `@iota/iota-sdk/utils`?

Thoralf-M: Currently don't see why we would have it there and not just in the IOTA-Names SDK

thibault-martinez: Alright

sdk/src/utils.ts
thibault-martinez: Please create an issue

thibault-martinez: https://github.com/iotaledger/iota-names/issues/60

thibault-martinez (approved):

PR #57, opened 2025-02-20
chore: Remove `controller` module

### LINKED ISSUE #37: Remove/rename `controller`

DaughterOfMars:
## Description

- Removes the `controller` module in `iota-names`.

Closes #37


packages/utils/sources/direct_setup.move
Thoralf-M, line 1: How is this rename related to the controller?

DaughterOfMars: It's not, but it's a stupid name so I changed it

Thoralf-M: Please keep such things separated, easier to review it when it's separated

DaughterOfMars: fine, done

Thoralf-M (approved):
There is also a mention of `Controller` in iota_names.move
```
/// The `App` type parameter is a witness which should be defined in the
/// original module (Controller, Registry, Registrar - whatever).
```

thibault-martinez:
As discussed on Slack, I don't think we should remove it in the end.

DaughterOfMars:
> As discussed on Slack, I don't think we should remove it in the end.

For what reason?

thibault-martinez:
Refer to https://github.com/iotaledger/iota-names/issues/37 for reasoning to close.

PR #68, opened 2025-02-24
fix: e2e tests, disable mainnet and testnet for now

Thoralf-M:
Use default `develop` branch instead of not existing `main` for localnet and disable mainnet and testnet for now as we don't have it deployed there
Fixes https://github.com/iotaledger/iota-names/issues/67

Tested by running `iota start --with-faucet --force-regenesis --epoch-duration-ms 120000` and `pnpm test:e2e`

PR #69, opened 2025-02-24
fix: use framework/devnet branch for localnet deployment

Alex6323:
current still set to `main`

### DIFF
--- a/scripts/init/packages.ts
+++ b/scripts/init/packages.ts
@@ -37,7 +37,7 @@ const parseCreatedObject = (data: IotaTransactionBlockResponse, objectType: stri
 };
 
 export const Packages = (network: Network) => {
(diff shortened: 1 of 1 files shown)


scripts/init/packages.ts
```diff
@@ -37,7 +37,7 @@ const parseCreatedObject = (data: IotaTransactionBlockResponse, objectType: stri
 };
 
 export const Packages = (network: Network) => {
-	const rev = network === 'localnet' ? 'main' : `framework/${network}`;
+	const rev = network === 'localnet' ? 'framework/devnet' : `framework/${network}`;
```
Thoralf-M, line 40: Why would you go for framework/devnet? `main` was there default branch, for us the equivalent would be `develop` I also changed this in https://github.com/iotaledger/iota-names/pull/68 as the test was failing otherwise, but I can also remove my change there

PR #72, opened 2025-02-24
feat: `domain::parent` returns `Option<Domain>`

### LINKED ISSUE #10: Make return type of `iotans::domain::parent` optional
Currently its returning `Domain`, but should be optional instead as it could return something invalid otherwise (no label).

Alex6323:
Closes #10


Thoralf-M (changes requested):
Tests need to be updated

Thoralf-M:
Best to wait until https://github.com/iotaledger/iota-names/pull/75 is merged

Thoralf-M (approved):

thibault-martinez (approved):

PR #74, opened 2025-02-26
Pull upstream changes

DaughterOfMars:
## Description

Pulls upstream changes which introduced payments using other coins. Most of that has been excised, but there were structural changes to pull as well as the removal of the SDK.


documentation/pages/developer/sdk/transactions.mdx
DaughterOfMars, line 1: FYI these files are pretty wrong but I figure we'll go back and re-do the docs later.

.github/workflows/iota-names-build-tx.yaml
```diff
@@ -48,6 +49,11 @@ jobs:
       - name: Install IOTA using Homebrew
         run: brew install iotaledger/tap/iota
 
+      - name: YAML Setup
+        run: |
+          sui client --yes new-env --rpc https://fullnode.mainnet.sui.io:443 --alias mainnet
+          sui client switch --env mainnet
```
Thoralf-M: Needs to be changed

.github/workflows/typescript.yaml
```diff
@@ -68,10 +68,3 @@ jobs:
       - name: Install Dependencies
         run: |
           pnpm install
-
-      - name: Prepare local network & run SDK tests
-        env:
-          IS_CI_JOB: true
-          NETWORK: localnet
-        run: |
-          iota start --with-faucet --force-regenesis --epoch-duration-ms 120000 > /dev/null 2>&1 & cd sdk && VITE_IOTA_BIN="iota" pnpm test:e2e
```
Thoralf-M, line 77: I wouldn't remove this, AFAIK they only removed localnet because of the pyth package which they only have published in their testnet and mainnet

sdk/README.md
Thoralf-M, line 1: They removed the SDK because they moved it to another repo, we don't have this yet, so I would keep it here until we have moved it somewhere else

.github/workflows/iota-names-build-tx.yaml
DaughterOfMars: thanks

.github/workflows/typescript.yaml
DaughterOfMars: No, they removed it because they removed the SDK

sdk/README.md
DaughterOfMars: Yeah but we're not going to be able to pull upstream changes to match these if we just keep the old code

scripts/transactions/mainnet-setup.ts
```diff
@@ -0,0 +1,107 @@
...
+	authorizeApp,
+	newPriceConfig,
+	newRenewalConfig,
+	addConfig,
+} from '../init/authorization';
+import { prepareMultisigTx } from '../utils/utils';
+import { NANOS_PER_IOTA } from '@iota/iota-sdk/utils';
+
+// Upgrade Suins
+const setupSuins = (txb: Transaction) => {
```
Thoralf-M, line 18: Needs to be updated

.github/workflows/typescript.yaml
Thoralf-M: Hm okay, but we should still have such a test

sdk/README.md
Thoralf-M: I thought it was also used in the scripts, but if not I'm also fine to have it removed for now as we don't need it then

DaughterOfMars:
Closed in favor of #75

PR #80, opened 2025-03-18
feat: Enable managing packages after initial publish

### LINKED ISSUE #17: Find a way to manage packages after initial publish
The `managePackage` typescript function which is used by the publish scripts should handle packages after the initial publishing. Currently it assumes the latest package ID is the same as the initial one and the version is 1.

DaughterOfMars:
## Description

Enables package management after the initial publish by parsing the lock file.

Closes #17


Thoralf-M:
Test fails with
```
/home/runner/work/iota-names/iota-names/scripts/utils/utils.ts:297
		return env[activeEnv] as JsonMap;
            ^
TypeError: Cannot read properties of undefined (reading 'None')
```
And if you run `pnpm run lint:fix` it should fix the TS issues

Thoralf-M (approved):

scripts/utils/utils.ts
```diff
@@ -67,24 +68,38 @@ export const publishPackage = (txb: Transaction, path: string, configPath?: stri
...
+	const activeEnv = getActiveEnv(configPath);
+
+	var versionNumber = '1';
+	var originalId = packageId;
+	var latestId = packageId;
+	const metadata = getManagedMetadata(packageFolder, activeEnv);
+	var chainId: string;
+	if (metadata) {
+		chainId = metadata['chain-id'] as string;
+		versionNumber = String((metadata['published-version'] as number) + 1);
```
thibault-martinez, line 81: Why always `+ 1`? Couldn't `managePackage` not be used in a non-publish context?

thibault-martinez: @Thoralf-M do you have an answer to this?

Thoralf-M: > Couldn't managePackage not be used in a non-publish context?

It could, but currently it isn't and since it's not a library but just in the scripts it should be fine this way
If we ever want to use it for something else we have to adjust it then

thibault-martinez (approved):

PR #84, opened 2025-03-24
feat: add auction package

### LINKED ISSUE #81: Implement auction
> Auction vs. Direct Purchase
> 
> I agree that in the initial phase, an auction-based system should be implemented.
> A successful purchase should require 24 hours without competing bids.
> If a bid is placed, the auction end time should be extended by 10 minutes (or up to 1 hour).
> The system should not allow rapid bidding extensions within just 10 seconds.

Take auction back from upstream (we removed it).
We can maybe make it its own package instead of module of the main package.

Thoralf-M:
Add an auction package to make registration fairer at the start

Closes #81 

Tested in the devnet to start and bid, bid and claim

Also tested authorization and deauthorization:
```
# authorize auction app 
IOTA_NAMES_PACKAGE_ID=0xa1d2ed2008d31d358cfaf61a89aa7cfaa78ed183dbe683620258e98c59f48b13
IOTA_NAMES_OBJECT_ID=0xbf3563622035af599057c46f4b871e0a9817c7bab759354532402be6d9538ba3
ADMIN_CAP=0x3014281d985cf1e2f640884d905c6c81aeb5951d14482e701428d4e611404a68
AUCTION_PACKAGE_ID=0xe24e72f0623ea19b4b2ec847e2b208033c6f7938b50e31efe4996aa2f23d4477
AUCTION_HOUSE_OBJECT_ID=0x85f493ba298b68af3e4812385460e21ddc5aa61273efd9dc54aa6919848090e4
iota client ptb \
--move-call $IOTA_NAMES_PACKAGE_ID::iota_names::authorize_app "<$AUCTION_PACKAGE_ID::auction::App>" @$ADMIN_CAP @$IOTA_NAMES_OBJECT_ID \
--dry-run
# deauthorize auction app
iota client ptb \
--move-call $IOTA_NAMES_PACKAGE_ID::iota_names::deauthorize_app "<$AUCTION_PACKAGE_ID::auction::App>" @$ADMIN_CAP @$IOTA_NAMES_OBJECT_ID \
--dry-run
```
Started an auction
```
iota client ptb \
--split-coins gas "[100000000]" \
--assign payment_coins \
--move-call $AUCTION_PACKAGE_ID::auction::start_auction_and_place_bid @$AUCTION_HOUSE_OBJECT_ID @$IOTA_NAMES_OBJECT_ID '"auction.iota"' payment_coins.0 @0x6 \
--dry-run
```

Also tried to finalize auctions 
```
ADMIN_CAP=0x3014281d985cf1e2f640884d905c6c81aeb5951d14482e701428d4e611404a68
AUCTION_PACKAGE_ID=0xe24e72f0623ea19b4b2ec847e2b208033c6f7938b50e31efe4996aa2f23d4477
AUCTION_HOUSE_OBJECT_ID=0x85f493ba298b68af3e4812385460e21ddc5aa61273efd9dc54aa6919848090e4
iota client ptb \
--move-call $AUCTION_PACKAGE_ID::auction::admin_try_finalize_auctions @$ADMIN_CAP @$AUCTION_HOUSE_OBJECT_ID 100 @0x6 \
--dry-run
(without dry-run: Transaction Digest: G79rmSw6PsYWLUm342BAvrBGSqQawWTPznVChs4oAwjp)
```

The error messages are visible in the output like this:
```
Error executing transaction '882EAM6EywWDU5R5UUr15PDcgNnFqfYxB4fY9xvj6iYu': 2nd command aborted within function '0xe24e72f0623ea19b4b2ec847e2b208033c6f7938b50e31efe4996aa2f23d4477::auction::start_auction_and_place_bid' at line 107. Aborted with 'EInvalidBidValue' -- 'The bid value is too low (compared to min_bid or previous bid).'
```


scripts/init/packages.ts
```diff
@@ -118,9 +118,9 @@ export const Packages = (network: string) => {
 							[5, 63],
 						],
 						prices: [
-							150 * Number(NANOS_PER_IOTA),
-							50 * Number(NANOS_PER_IOTA),
-							5 * Number(NANOS_PER_IOTA),
+							0.15 * Number(NANOS_PER_IOTA),
+							0.05 * Number(NANOS_PER_IOTA),
+							0.005 * Number(NANOS_PER_IOTA),
```
thibault-martinez, line 123: Maybe an issue to track reverting this?

Thoralf-M: Wouldn't we want to set something different instead of reverting?

thibault-martinez: Yeah, sure

Thoralf-M: https://github.com/iotaledger/iota-names/issues/124
But since we can replace the configs at any point it doesn't really matter

thibault-martinez:
CI is failing, feel free to bring back `nanos_per_iota`

packages/auction/sources/auction.move
```diff
@@ -0,0 +1,446 @@
...
+
+/// The AuctionHouse application.
+public struct AuctionHouse has key, store {
+    id: UID,
+    balance: Balance<IOTA>,
+    auctions: LinkedTable<Domain, Auction>,
+}
+
+/// The Auction application.
+#[allow(lint(coin_field))]
```
thibault-martinez, line 71: Do you know what's that for? Still raised?

packages/auction/sources/auction.move
```diff
@@ -0,0 +1,446 @@
...
+    // left.
+    // Auctions can't be finished until there is at least
+    // `AUCTION_MIN_QUIET_PERIOD_MS`
+    // time where there are no bids.
+    if (end_timestamp_ms - clock.timestamp_ms() < AUCTION_MIN_QUIET_PERIOD_MS) {
+        let new_end_timestamp_ms = clock.timestamp_ms() + AUCTION_MIN_QUIET_PERIOD_MS;
+
+        // Only extend the auction if the new auction end time is before
+        // the NFT's expiration timestamp
+        if (new_end_timestamp_ms < nft.expiration_timestamp_ms()) {
```
thibault-martinez, line 191: Does that mean that an auction could technically run for an entire year?

packages/auction/sources/auction.move
```diff
@@ -0,0 +1,446 @@
...
+    event::emit(AuctionFinalizedEvent {
+        domain,
+        start_timestamp_ms,
+        end_timestamp_ms,
+        winning_bid: coin::value(&current_bid),
+        winner,
+    });
+
+    // Extract the NFT and their bid, returning the NFT to the user
+    // and sending the proceeds of the auction to iota_names
```
thibault-martinez: to iota_names?

packages/auction/sources/auction.move
```diff
@@ -0,0 +1,446 @@
...
+            some(auction.start_timestamp_ms),
+            some(auction.end_timestamp_ms),
+            some(auction.winner),
+            some(highest_amount),
+        )
+    };
+    (none(), none(), none(), none())
+}
+
+public fun collect_winning_auction_fund(
```
thibault-martinez, line 285: Not sure I understand this function

packages/auction/sources/auction.move
```diff
@@ -0,0 +1,446 @@
...
+///
+/// This will attempt to do as much as possible of the following
+/// based on the provided `operation_limit`:
+/// - claim the winning bid and place in `AuctionHouse.balance`
+/// - push the `IotaNamesRegistration` to the winner
+/// - push loosing bids back to their respective account owners
+///
+/// Once all of the above has been done the auction is destroyed,
+/// freeing on-chain storage.
+public fun admin_finalize_auction(
```
thibault-martinez, line 326: Why would admin do that?

packages/auction/sources/auction.move
Thoralf-M: Sub-optimal 'iota::coin::Coin' field type. Using 'iota::balance::Balance' instead will be more space efficient

packages/auction/sources/auction.move
Thoralf-M: Yes

packages/auction/sources/auction.move
thibault-martinez: Do you think they added the lint later because it was too late to update it? Should we do that in the first place?

packages/auction/sources/auction.move
Thoralf-M: This just moves the bid amount to the auction house, so one can then transfer it using `admin_withdraw_funds()`
It's similar to `admin_finalize_auction()` just with the difference that the winner still needs to call `claim()` here

packages/auction/sources/auction.move
Thoralf-M: To get the funds like `collect_winning_auction_fund()` but also help users so they don't have to claim the NFT in an extra tx

packages/auction/sources/auction.move
Thoralf-M: Was also weird in upstream, changed it a bit

packages/auction/sources/auction.move
```diff
@@ -0,0 +1,446 @@
...
+    clock: &Clock,
+    ctx: &mut TxContext,
+) {
+    let domain = domain::new(domain_name);
+    let auction = &mut self.auctions[domain];
+    // Ensure that the auction is over
+    assert!(clock.timestamp_ms() > auction.end_timestamp_ms, EAuctionNotEndedYet);
+
+    let amount = auction.current_bid.value();
+    self.balance.join(auction.current_bid.split(amount, ctx).into_balance());
```
thibault-martinez, line 297: Do we really need that split here?

packages/auction/sources/auction.move
thibault-martinez: Alright, this feels tedious but maybe they did it for very big auction bids

packages/auction/sources/auction.move
Thoralf-M: Since this field is only used for the time of an Auction I don't think it doesn't really matter. If someone places a bid, we would then need to create this `Coin` object so we can transfer the balance to the previous winner because one can't just transfer a `Balance`

packages/auction/sources/auction.move
Thoralf-M: Yes, because we don't destroy the auction we still have to keep the coin object, even if it's empty

thibault-martinez: Thanks

packages/auction/sources/auction.move
thibault-martinez: By "help" you mean not spend gas?

thibault-martinez (approved):

packages/auction/sources/auction.move
Thoralf-M: More like not have to go to some dapp and sign a tx to call this function, but yes they also don't have to spend the gas if the admin does it

thibault-martinez: I didn't see this as being annoying since they already bid on the dApp but ok, fair enough

DaughterOfMars (approved):

Alex6323 (approved):

PR #91, opened 2025-03-31
feat(iota-names)!: remove consts and public funs that aren't required

### LINKED ISSUE #56: Why is MAX_DOMAIN_LENGTH  == 63?
Why exactly 63?
```
/// The maximum length of a domain name.
const MAX_DOMAIN_LENGTH: u8 = 63;
```
And in TS, why is there `const MAX_IOTA_NAME_LENGTH = 235;`?

Thoralf-M:
Remove consts and public funs that aren't required

Fixes https://github.com/iotaledger/iota-names/issues/56

### DIFF
--- a/packages/iota-names/sources/iota_names.move
+++ b/packages/iota-names/sources/iota_names.move
@@ -227,8 +227,6 @@ public struct Test has drop {}
 
 #[test_only]
 const NANOS_PER_IOTA: u64 = 1_000_000_000;
-#[test_only]
-public fun nanos_per_iota(): u64 { NANOS_PER_IOTA }
 
 #[test_only]
 public fun new_for_testing(ctx: &mut TxContext): (IotaNames, AdminCap) {
@@ -263,9 +261,9 @@ public fun new_pricing_config(): PricingConfig {
     let range2 = new_range(vector[4, 4]);
     let range3 = new_range(vector[5, 63]);
     let prices = vector[
-        1200 * nanos_per_iota(),
-        200 * nanos_per_iota(),
-        50 * nanos_per_iota(),
+        1200 * NANOS_PER_IOTA,
+        200 * NANOS_PER_IOTA,
+        50 * NANOS_PER_IOTA,
     ];
 
     pricing_config::new(
--- a/packages/iota-names/sources/iota_names.move
+++ b/packages/iota-names/sources/iota_names.move
@@ -226,6 +226,11 @@ use iota_names::pricing_config::{Self, new_range, PricingConfig};
 public struct Test has drop {}
 
 #[test_only]
+const NANOS_PER_IOTA: u64 = 1_000_000_000;
+#[test_only]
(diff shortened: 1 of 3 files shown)


packages/iota-names/tests/controller_tests.move
```diff
@@ -13,14 +13,14 @@ use iota::iota::IOTA;
 use iota::test_scenario::{Self, Scenario, ctx};
 use iota::test_utils::{assert_eq, destroy};
 use iota::vec_map::VecMap;
-use iota_names::constants::{nanos_per_iota, year_ms};
+use iota_names::constants::{ year_ms};
```
thibault-martinez:
```suggestion
use iota_names::constants::{year_ms};
```

packages/iota-names/sources/iota_names.move
```diff
@@ -225,6 +225,11 @@ use iota_names::pricing_config::{Self, new_range, PricingConfig};
 #[test_only]
 public struct Test has drop {}
 
+#[test_only]
+const NANOS_PER_IOTA: u64 = 1_000_000_000;
+#[test_only]
+public fun nanos_per_iota(): u64 { NANOS_PER_IOTA }
```
thibault-martinez: Maybe we should at least keep this one idk

thibault-martinez (approved):

packages/iota-names/sources/iota_names.move
thibault-martinez: And if we want it here, I don't think we need the function, just the const is enough since it's not an API

Thoralf-M: Keep it for what if it's not used and if you can't even access it from the outside?

thibault-martinez: > you can't even access it from the outside?

What do you mean?

packages/iota-names/tests/controller_tests.move
```diff
@@ -13,14 +13,14 @@ use iota::iota::IOTA;
 use iota::test_scenario::{Self, Scenario, ctx};
 use iota::test_utils::{assert_eq, destroy};
 use iota::vec_map::VecMap;
-use iota_names::constants::{nanos_per_iota, year_ms};
+use iota_names::constants::{year_ms};
```
Alex6323:
```suggestion
use iota_names::constants::year_ms;
```

packages/iota-names/sources/iota_names.move
Thoralf-M: Functions you can call from other packages, consts not

thibault-martinez: Yeah I know, when I said "keep it" I meant keep the const + its function.
Anyway, it's a very basic const that people may end up needing but fine. I would still remove the function here though.

Alex6323 (approved):

thibault-martinez (approved):

PR #75, opened 2025-02-27
Pull Upstream Payment and SDK Changes

DaughterOfMars:
## Description

Another attempt to pull upstream changes, but this time keeping the payments package. I have removed anything `pyth` related and simplified the API.

PR #97, opened 2025-04-02
feat(workflows): test publishing

Thoralf-M:
Test publishing the package and registering a name with it, testing this was removed in https://github.com/iotaledger/iota-names/pull/75 as in upstream they don't publish to a localnet anymore because of the pyth dependency


.github/workflows/typescript.yaml
```diff
@@ -74,4 +74,14 @@ jobs:
           IS_CI_JOB: true
           NETWORK: localnet
         run: |
-          iota start --with-faucet --force-regenesis --epoch-duration-ms 120000 > /dev/null 2>&1 & cd sdk && VITE_IOTA_BIN="iota" pnpm test:e2e
```
thibault-martinez, line 77: How was this line working with the `cd` just after `iota start`?

Thoralf-M: Because of the `&` it just runs everything before in the background and then starts from the `cd sdk`

thibault-martinez: Hmm ok, I thought we still needed a connector of some sort after the background operator, TIL 👍🏻

.github/workflows/typescript.yaml
```diff
@@ -74,4 +74,14 @@ jobs:
...
+          iota start --with-faucet --force-regenesis --epoch-duration-ms 120000 > /dev/null 2>&1 &
+          cd sdk
+          VITE_IOTA_BIN="iota" pnpm test:e2e
+          sleep 5 # wait some time for the network
+          iota client --yes new-env --alias localnet --rpc http://127.0.0.1:9000
+          iota client switch --env localnet
+          iota client faucet
+          cd ../scripts
+          export NETWORK=localnet
+          pnpm ts-node init/init.ts
```
thibault-martinez: Will need to update this after https://github.com/iotaledger/iota-names/pull/101

scripts/transactions/test-register-name.ts
thibault-martinez, line 1: Is this the right place to do tests? Feels a bit weird

Thoralf-M: I put it there because I thought one could also run this to test something manually locally, but also wasn't sure where to put it, any suggestion?

.github/workflows/typescript.yaml
Thoralf-M: Done in https://github.com/iotaledger/iota-names/pull/97/commits/f2179b0f739b9574e6ee693186dcc1fd427ab554

scripts/transactions/test-register-name.ts
thibault-martinez: `scripts/tests/*`?

Thoralf-M: If we want it only as test yes, but as I said I think it might be useful not as "test" like in the CI only, but can only be useful to run locally so you have a registered name and then `scripts/tests/*` would be also wrong
Or do you think it's not that useful to have as script to run manually and should be only test for the CI?

scripts/transactions/test-register-name.ts
```diff
@@ -0,0 +1,103 @@
+// Copyright (c) Mysten Labs, Inc.
```
thibault-martinez: Did we copy anything?

scripts/transactions/test-register-name.ts
thibault-martinez: IMO we shouldn't care too much about use cases here, we're not writing an SDK

scripts/transactions/test-register-name.ts
Thoralf-M: Some individual parts, but I guess not needed for this file

Thoralf-M: Changed in https://github.com/iotaledger/iota-names/pull/97/commits/edfd360a0481a3cf2c75b6dacb34a99720c84131

scripts/transactions/test-register-name.ts
Thoralf-M: Moved in https://github.com/iotaledger/iota-names/pull/97/commits/edfd360a0481a3cf2c75b6dacb34a99720c84131

scripts/transactions/test-register-name.ts
thibault-martinez: Doesn't look changed to me. But also fine keeping it.

scripts/tests/test-register-name.ts
thibault-martinez, line 1: I'd drop the `test-` prefix then

scripts/transactions/test-register-name.ts
Thoralf-M: Ah I changed it and then run pnpm lint:fix which adds it again

scripts/tests/register-name.ts
```diff
@@ -0,0 +1,103 @@
...
+		target: `${config.packageId}::payment::register`,
+		arguments: [receipt, tx.object(config.iotaNames), tx.object('0x6')],
+	});
+	const signer = getSigner();
+	tx.transferObjects([nft], tx.pure.address(signer.getPublicKey().toIotaAddress()));
+
+	const res = await signAndExecute(tx, network);
+	if (res.errors) {
+		console.error('Transaction failed:', res.errors);
+		return;
```
thibault-martinez: Would this actually make the CI fail or would it just log ?

Thoralf-M: This would just log, changing to throw

Thoralf-M: https://github.com/iotaledger/iota-names/pull/97/commits/01b0e119f9dec9cf353f9f935d758e9e1519a869

thibault-martinez (approved):

DaughterOfMars (approved):

PR #101, opened 2025-04-02
Pass network as CLI argument to init.ts

### LINKED ISSUE #100: Pass network as argument to init.ts instead of env variable

theiari:
Pass network as CLI argument to init.ts instead of relying on the env variable, makes it easier to use


scripts/init/init.ts
```diff
@@ -5,15 +5,20 @@
 import { publishPackages } from './publish';
 import { setup } from './setup';
 
+// Extract network argument from command-line arguments
+const args = process.argv.slice(2); // Get arguments passed after `node init.ts`
+const network = args[0]; // First argument should be the network
```
thibault-martinez, line 16: Maybe we should check the length of `args` and warn here if it's not correct, because otherwise it'll panic and the check L13 is useless anyway?

scripts/init/init.ts
```diff
@@ -5,15 +5,20 @@
...
+
 export const init = async (network: string | undefined, isCIJob: boolean) => {
-	if (!network)
+	if (!network) {
 		throw new Error(
-			'Network not defined. Please run `export NETWORK=mainnet|testnet|devnet|localnet`',
+			'Network not defined. Please run `node init.js <network>` (e.g., mainnet, testnet, devnet, localnet)',
```
Thoralf-M: It's easier to use `ts-node` to execute .ts files as this will run the `tsc` step for you and since not everyone wants to install it globally it can be used via a script in package.json or with `pnpm` as in the README.md: `pnpm ts-node init/init.ts`

scripts/init/init.ts
```diff
@@ -5,15 +5,20 @@
 import { publishPackages } from './publish';
 import { setup } from './setup';
 
+// Extract network argument from command-line arguments
+const args = process.argv.slice(2); // Get arguments passed after `node init.ts`
+const network = args[0]; // First argument should be the network
+
 export const init = async (network: string | undefined, isCIJob: boolean) => {
-	if (!network)
+	if (!network || args.length !== 1) {
```
thibault-martinez: Not sure we can actually check here, by the time we call `init`, L10 would have already panicked

scripts/init/init.ts
```diff
@@ -5,15 +5,26 @@
...
+}
+
+const network = args[0]; // First argument should be the network
+
 export const init = async (network: string | undefined, isCIJob: boolean) => {
-	if (!network)
+	if (!network) {
 		throw new Error(
-			'Network not defined. Please run `export NETWORK=mainnet|testnet|devnet|localnet`',
+			'Network not defined or too many arguments. Please run `pnpm ts-node init.js <network>` (e.g., mainnet, testnet, devnet, localnet)',
```
thibault-martinez:
```suggestion
			'Network not defined. Please run `pnpm ts-node init.js <network>` (e.g., mainnet, testnet, devnet, localnet)',
```

scripts/init/init.ts
```diff
@@ -5,15 +5,26 @@
...
+}
+
+const network = args[0]; // First argument should be the network
+
 export const init = async (network: string | undefined, isCIJob: boolean) => {
-	if (!network)
+	if (!network) {
 		throw new Error(
-			'Network not defined. Please run `export NETWORK=mainnet|testnet|devnet|localnet`',
+			'Network not defined. Please run `pnpm ts-node init.js <network>` (e.g., mainnet, testnet, devnet, localnet)',
```
thibault-martinez:
```suggestion
			'Network not defined. Please run `pnpm ts-node init.ts <network>` (e.g., mainnet, testnet, devnet, localnet)',
```

Thoralf-M (approved):

thibault-martinez (approved):

PR #104, opened 2025-04-02
Transfer to new owner in init.ts

Alex6323:


scripts/init/init.ts
```diff
@@ -2,18 +2,53 @@
...
+
+    const client = getClient(network);
+    const res = await client.getOwnedObjects({
+        newOwner: getActiveAddress(),
+        options: { showType: true },
+    });
+    const ownedNonCoinObjects = (res.data as IotaObjectResponse[])
+        .filter((x) => x.data?.type !== '0x2::coin::Coin<0x2::iota::IOTA>')
+        .map((x) => x.data?.objectId)
+        .filter((id): id is string => typeof id === "string");
```
Thoralf-M: Wait, isn't this dangerous? :trollface:

Thoralf-M: I found a solution that doesn't require hardcoded values but still filters it out:
```TS
	const config = readPackageInfo(network);
	const configValues = Object.entries(config)
		.map(([key, value]) => value)
		.filter((v) => typeof v === 'string');

	const ownedObjectsPage = await client.getOwnedObjects({
		owner: config.adminAddress,
		options: { showContent: true },
	});

	let objectsToTransfer = [];
	for (const object of ownedObjectsPage.data) {
		for (const configValue of configValues) {
			if (JSON.stringify(object).includes(configValue)) {
				objectsToTransfer.push(object.data?.objectId);
				break;
			}
		}
	}
	console.log('Objects to transfer:', objectsToTransfer.length);
	console.log(objectsToTransfer);
```

Thoralf-M: But it requires another small fix, to also have the denylist packages included in the config, it's currently missing https://github.com/iotaledger/iota-names/pull/106

Alex6323: Looks a bit less dangerous :trollface:

scripts/init/init.ts
```diff
@@ -2,18 +2,57 @@
...
+		.filter((x) => x.data?.type !== '0x2::coin::Coin<0x2::iota::IOTA>')
+		.map((x) => x.data?.objectId)
+		.filter((id): id is string => typeof id === 'string');
+
+	const tx = new Transaction();
+	tx.transferObjects(ownedNonCoinObjects, newOwner);
+
+	const result = await signAndExecute(tx, network);
+	const response = await client.waitForTransaction({ digest: result.digest });
+	console.log(`Transaction digest: ${response.digest}`);
```
Thoralf-M:
```suggestion
	const result = await signAndExecute(tx, network);
	await client.waitForTransaction({ digest: result.digest });
	console.log(`Transaction digest: ${result.digest}`);
```

scripts/init/init.ts
```diff
@@ -2,18 +2,68 @@
...
+) => {
 	if (!network)
 		throw new Error(
-			'Network not defined. Please run `export NETWORK=mainnet|testnet|devnet|localnet`',
+			'`network` not defined. Please run `export NETWORK=mainnet|testnet|devnet|localnet`',
+		);
+
+	if (!newOwner)
+		throw new Error(
+			'`newOwner` not defined. Please provide the new owner address of IOTA-Names, e.g. a multisig address',
```
Thoralf-M: I think the newOwner should be optional, if not provided, just don't transfer anything at the end
Makes it easier for testing

scripts/init/init.ts
```diff
@@ -2,29 +2,70 @@
...
 
+import { Transaction } from '@iota/iota-sdk/transactions';
+
+import { readPackageInfo } from '../config/constants';
+import { getClient, signAndExecute } from '../utils/utils';
 import { publishPackages } from './publish';
 import { setup } from './setup';
 
 // Extract network argument from command-line arguments
 const args = process.argv.slice(2); // Get arguments passed after `node init.ts`
```
thibault-martinez: `node init.ts` is a bit weird btw, not really the command

scripts/init/init.ts
```diff
@@ -2,29 +2,70 @@
...
 // SPDX-License-Identifier: Apache-2.0
 
+import { Transaction } from '@iota/iota-sdk/transactions';
+
+import { readPackageInfo } from '../config/constants';
+import { getClient, signAndExecute } from '../utils/utils';
 import { publishPackages } from './publish';
 import { setup } from './setup';
 
 // Extract network argument from command-line arguments
```
thibault-martinez: Also mention newOwner?

scripts/init/init.ts
Alex6323: done

scripts/init/init.ts
Alex6323: yeah, I copied that

scripts/init/init.ts
```diff
@@ -2,29 +2,70 @@
...
 const args = process.argv.slice(2); // Get arguments passed after `node init.ts`
 if (args.length !== 1) {
```
thibault-martinez: It could be 2 and still be fine?

scripts/init/init.ts
thibault-martinez: I know, but let's change it pls

Alex6323: removed

scripts/init/init.ts
Alex6323: done

scripts/init/init.ts
Alex6323: fixed

Thoralf-M (approved):

thibault-martinez (approved):

PR #120, opened 2025-04-03
fix(scripts): fix-funds-to-treasury

Thoralf-M:
The funds are in a dynamic field now, also the folder needs to be created first if it doesn't exist yet

### DIFF
--- a/scripts/utils/utils.ts
+++ b/scripts/utils/utils.ts
@@ -231,6 +231,10 @@ export const prepareMultisigTx = async (tx: Transaction, network: string, addres
 		const output_location =
 			process.env.NODE_ENV === 'development' ? './tx/tx-data-local.txt' : './tx/tx-data.txt';
 
+		const dir = path.dirname(output_location);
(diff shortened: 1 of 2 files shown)


scripts/utils/utils.ts
```diff
@@ -231,6 +231,10 @@ export const prepareMultisigTx = async (tx: Transaction, network: string, addres
 		const output_location =
 			process.env.NODE_ENV === 'development' ? './tx/tx-data-local.txt' : './tx/tx-data.txt';
 
+		const dir = path.dirname(output_location);
+		if (!fs.existsSync(dir)) {
+			fs.mkdirSync(dir, { recursive: true });
+		}
```
thibault-martinez, line 237: Why is this necessary?

Thoralf-M: Without it failed for me because the folder didn't exist

`ENOENT: no such file or directory`

thibault-martinez (approved):

PR #126, opened 2025-04-08
feat(iota-names)!: remove IotaNames::balance

### LINKED ISSUE #113: Remove IotaNames::balance
## Description

The payments are now just added as dynamic field like this:
```
{
            "name": {
              "type": {
                "repr": "0x9d1f612a3a4e7fdc886cc5fd23202c3b00b49044a21caa5622ff28c98e257646::iota_names::BalanceKey<0x0000000000000000000000000000000000000000000000000000000000000002::iota::IOTA>"
              }
            },
            "value": {
              "json": {
                "value": "10000000000"
              }
            }
          }
```
so we can probably remove the struct field so it's less confusing as it would always stay at 0 even though there is balance

Thoralf-M:
Remove the IotaNames::balance field as balances are stored as dynamic field

Fixes https://github.com/iotaledger/iota-names/issues/113

Should we remove the `_custom` from `app_add_custom_balance`, `withdraw_custom` now?

### DIFF
--- a/packages/iota-names/sources/iota_names.move
+++ b/packages/iota-names/sources/iota_names.move
@@ -97,7 +97,7 @@ fun init(otw: IOTA_NAMES, ctx: &mut TxContext) {
 // === Admin actions ===
 
 /// Withdraw from the IotaNames balance of a custom coin type.
-public fun withdraw_custom<T>(self: &mut IotaNames, _: &AdminCap, ctx: &mut TxContext): Coin<T> {
+public fun withdraw<T>(self: &mut IotaNames, _: &AdminCap, ctx: &mut TxContext): Coin<T> {
     let balance_key = BalanceKey<T> {};
     assert!(self.id.exists_(balance_key), ENoProfitsInCoinType);
 
@@ -137,7 +137,7 @@ public fun assert_app_is_authorized<App: drop>(self: &IotaNames) {
 // === Protected features ===
 
 /// Adds a balance of type `T` to the IotaNames protocol as an authorized app.
-public fun app_add_custom_balance<App: drop, T>(self: &mut IotaNames, _: App, balance: Balance<T>) {
+public fun app_add_balance<App: drop, T>(self: &mut IotaNames, _: App, balance: Balance<T>) {
     self.assert_app_is_authorized<App>();
     let key = BalanceKey<T> {};
     if (self.id.exists_(key)) {
--- a/packages/iota-names/sources/iota_names.move
+++ b/packages/iota-names/sources/iota_names.move
@@ -96,7 +96,7 @@ fun init(otw: IOTA_NAMES, ctx: &mut TxContext) {
 
 // === Admin actions ===
 
-/// Withdraw from the IotaNames balance of a custom coin type.
+/// Withdraw from the IotaNames balance of a provided coin type.
 public fun withdraw<T>(self: &mut IotaNames, _: &AdminCap, ctx: &mut TxContext): Coin<T> {
     let balance_key = BalanceKey<T> {};
     assert!(self.id.exists_(balance_key), ENoProfitsInCoinType);
--- a/packages/iota-names/sources/iota_names.move
+++ b/packages/iota-names/sources/iota_names.move
@@ -23,10 +23,9 @@
 /// the registry and the balance.
 module iota_names::iota_names;
 
-use iota::balance::{Self, Balance};
-use iota::coin::{Self, Coin};
+use iota::balance::Balance;
+use iota::coin::Coin;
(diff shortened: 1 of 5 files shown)


thibault-martinez:
> Should we remove the `_custom` from `app_add_custom_balance`, `withdraw_custom` now?

I would say so

packages/iota-names/sources/iota_names.move
```diff
@@ -93,26 +89,15 @@ fun init(otw: IOTA_NAMES, ctx: &mut TxContext) {
...
-/// and then send them somewhere specific or keep at the address.
-public fun withdraw(_: &AdminCap, self: &mut IotaNames, ctx: &mut TxContext): Coin<IOTA> {
-    let amount = self.balance.value();
-    assert!(amount > 0, ENoProfits);
-    coin::take(&mut self.balance, amount, ctx)
-}
-
 /// Withdraw from the IotaNames balance of a custom coin type.
-public fun withdraw_custom<T>(self: &mut IotaNames, _: &AdminCap, ctx: &mut TxContext): Coin<T> {
+public fun withdraw<T>(self: &mut IotaNames, _: &AdminCap, ctx: &mut TxContext): Coin<T> {
```
thibault-martinez, line 100: Wondering: the previous `withdraw` was asserting if the balance is 0, is that not something we'd do here as well?

Thoralf-M: I guess it would make sense to pay less/now gas for such case :+1:

Thoralf-M: Changed in https://github.com/iotaledger/iota-names/pull/126/commits/47e250b7d09649bcb6e8d9826004bb029ed8369c

packages/iota-names/sources/iota_names.move
```diff
@@ -93,30 +89,21 @@ fun init(otw: IOTA_NAMES, ctx: &mut TxContext) {
...
     let balance_key = BalanceKey<T> {};
     assert!(self.id.exists_(balance_key), ENoProfitsInCoinType);
```
thibault-martinez, line 102: Could rename this to simply `ENoProfits` if we wanted

Thoralf-M: Yes, but I choose not to as there could be profits, just for a different coin type (if we allow other coin types in the future)

thibault-martinez: Fair enough

packages/iota-names/sources/iota_names.move
```diff
@@ -93,30 +89,21 @@ fun init(otw: IOTA_NAMES, ctx: &mut TxContext) {
...
-/// same
-/// transaction. This is useful for the admin to withdraw funds from the IotaNames
-/// and then send them somewhere specific or keep at the address.
-public fun withdraw(_: &AdminCap, self: &mut IotaNames, ctx: &mut TxContext): Coin<IOTA> {
-    let amount = self.balance.value();
-    assert!(amount > 0, ENoProfits);
-    coin::take(&mut self.balance, amount, ctx)
-}
-
 /// Withdraw from the IotaNames balance of a custom coin type.
```
thibault-martinez: I think I would also change the comment a bit, makes it seems IOTA is not part of this atm

packages/iota-names/tests/unit/iota_names_tests.move
```diff
@@ -52,15 +52,6 @@ fun registry_management() {
 
 public struct TestApp has drop {}
 
-#[test, expected_failure(abort_code = ::iota_names::iota_names::EAppNotAuthorized)]
-/// Only authorized applications can add balance to IotaNames.
-fun app_add_to_balance_fail() {
```
thibault-martinez, line 57: Why not keep it?

thibault-martinez (approved):

packages/iota-names/tests/unit/iota_names_tests.move
Thoralf-M: It wasn't compiling as it used the `app_add_balance` I removed, but I changed it now to use the version with the coin :+1:

Thoralf-M: https://github.com/iotaledger/iota-names/pull/126/commits/dff249a3dd7b13a20a398c98c3789faaef841bbb

packages/iota-names/sources/iota_names.move
Thoralf-M: https://github.com/iotaledger/iota-names/pull/126/commits/dff249a3dd7b13a20a398c98c3789faaef841bbb

DaughterOfMars (approved):

PR #129, opened 2025-04-09
feat: updated auction and auction-test to use minimum overbid of 1 IOTA

### LINKED ISSUE #127: Auction: set a minimum overbid of 1 IOTA
## Description

Set a minimum overbid of 1 IOTA for auctions, so they don't run that long just because someone adds 1 nano

So this part here needs to be changed https://github.com/iotaledger/iota-names/blob/2d190d2e2996346030b1f82f30dfcba6d0453254/packages/auction/sources/auction.move#L107
also adjust existing tests/add new ones if needed

theiari:


packages/auction/sources/auction.move
```diff
@@ -52,7 +53,7 @@ const ENotWinner: vector<u8> =
     b"Not winner of the auction.";
 #[error]
 const EBidAmountTooLow: vector<u8> =
-    b"The bid amount is too low.";
+    b"The bid amount is too low. Minimum overbid should be at least 1 IOTA.";
```
Alex6323: What's the difference between `EInvalidBidValue` and `EBidAmountTooLow`. Is there a difference between a `bid value` and a `bid amount`. If not, can we consolidate those two errors maybe?

Alternatively - if this error is specific to overbids - we could rename it to `EInvalidOverbidValue` to make it consistent with the other error. Wdyt?

packages/auction/tests/auction-tests.move
```diff
@@ -915,3 +915,64 @@ fun test_admin_collect_fund_even_if_auction_is_deauthorized() {
...
+
+#[test, expected_failure(abort_code = auction::EBidAmountTooLow)] 
+fun test_overbid_lower_than_one_iota_10000000_NANO() {
+    let mut scenario_val = test_init();
+    let scenario = &mut scenario_val;
+    start_auction_and_place_bid_util(
+        scenario,
+        FIRST_ADDRESS,
+        utf8(FIRST_DOMAIN_NAME),
+        100* NANOS_PER_IOTA,
```
Alex6323:
```suggestion
        100 * NANOS_PER_IOTA,
```

packages/auction/tests/auction-tests.move
```diff
@@ -915,3 +915,64 @@ fun test_admin_collect_fund_even_if_auction_is_deauthorized() {
...
+
+#[test, expected_failure(abort_code = auction::EBidAmountTooLow)] 
+fun test_overbid_lower_than_one_iota_100_NANOS() {
+    let mut scenario_val = test_init();
+    let scenario = &mut scenario_val;
+    start_auction_and_place_bid_util(
+        scenario,
+        FIRST_ADDRESS,
+        utf8(FIRST_DOMAIN_NAME),
+        100* NANOS_PER_IOTA,
```
Alex6323:
```suggestion
        100 * NANOS_PER_IOTA,
```

packages/auction/tests/auction-tests.move
```diff
@@ -915,3 +915,64 @@ fun test_admin_collect_fund_even_if_auction_is_deauthorized() {
...
+    );
+    place_bid_util(
+        scenario,
+        SECOND_ADDRESS,
+        utf8(FIRST_DOMAIN_NAME),
+        1,
```
Alex6323: not sure, but wouldn't  just one of those tests suffice to test for this error, or are those numbers picked specifically?

packages/auction/sources/auction.move
Thoralf-M: I think we should change `EInvalidBidValue`, as it says `(compared to min_bid or previous bid)` which can't be true for the place where it's used, it's always the first bid there and can only be lower than the price for the name

packages/auction/tests/auction-tests.move
theiari: Actually yes... I arbitrarily picked some numbers lower than 1 IOTA but one test is enough

packages/auction/sources/auction.move
theiari: Yes `EInvalidBidValue` can only be triggered by invalid (too low) starting bid. Previous bid is not related.

_and (?first bid?) can only be lower than the price for the name_ Isn't the opposite instead?
`assert!(bid.value() >= min_price, EInvalidBidValue);`
Rejects first bid lower than name price, or am I wrong?

Thoralf-M: @theiari yes, what I meant is that it can't return an error for something else than `first bid lower than name price`, because there can't be any bids before

Thoralf-M (approved):

packages/auction/tests/auction-tests.move
```diff
@@ -916,3 +916,23 @@ fun test_admin_collect_fund_even_if_auction_is_deauthorized() {
...
+        utf8(FIRST_DOMAIN_NAME),
+        100 * NANOS_PER_IOTA,
+    );
+    place_bid_util(
+        scenario,
+        SECOND_ADDRESS,
+        utf8(FIRST_DOMAIN_NAME),
+        100,
+        0,
+    );
```
thibault-martinez, line 976: This is actually not testing the under-overbid, maybe we can add more test cases

Thoralf-M: Multiple tests above do this `place_bid_util()`

thibault-martinez: Not really, there is only one (`test_place_bid_aborts_if_value_is_too_low`), that already tests underbid. We could test equality and under-overbid. This added test we already have.

thibault-martinez: (edit: overbid -> under-overbid as it wasn't super clear)

Thoralf-M: Ah yeah should just be something like `100 * NANOS_PER_IOTA+1`?

thibault-martinez: Yeah, I would actually have two tests for `100 * NANOS_PER_IOTA` and `100 * NANOS_PER_IOTA + 1`

Alex6323 (approved):

packages/auction/tests/auction-tests.move
```diff
@@ -916,3 +916,43 @@ fun test_admin_collect_fund_even_if_auction_is_deauthorized() {
...
+        SECOND_ADDRESS,
+        utf8(FIRST_DOMAIN_NAME),
+        100 * NANOS_PER_IOTA,
+        0,
+    );
+    scenario_val.end();
+}
+
+#[test, expected_failure(abort_code = auction::EBidAmountTooLow)]
+fun test_overbid_lower_than_one_iota_101_NANOS() {
```
thibault-martinez: Not exactly sure what `lower_than_one_iota_101_NANOS` means

packages/auction/tests/auction-tests.move
```diff
@@ -916,3 +916,43 @@ fun test_admin_collect_fund_even_if_auction_is_deauthorized() {
 
     scenario_val.end();
 }
+
+#[test, expected_failure(abort_code = auction::EBidAmountTooLow)] 
+fun test_overbid_lower_than_one_iota_100_NANOS() {
```
thibault-martinez: This one is not really lower anymore, it's equal

packages/auction/tests/auction-tests.move
Thoralf-M: Maybe rename it like this
```suggestion
fun test_lower_than_min_overbid() {
```

Thoralf-M (approved):

packages/auction/sources/auction.move
```diff
@@ -29,12 +29,13 @@ const DEFAULT_DURATION: u8 = 1;
...
 const AUCTION_MIN_QUIET_PERIOD_MS: u64 = 10 * 60 * 1000;
-
+/// The overbid must be at least of 1 IOTA, which is 10^9 NANOs
+const AUCTION_MIN_OVERBID_VALUE_IOTA: u64 = 1_000_000_000;
 // === Abort codes ===
 
 #[error]
 const EInvalidBidValue: vector<u8> =
-    b"The bid value is too low (compared to min_bid or previous bid).";
+    b"The bid value is too low (compared to min_bid).";
```
DaughterOfMars: There isn't actually any value called `min_bid` that the user will see so maybe just
```suggestion
    b"The bid value is too low (compared to minimum bid of 1 IOTA).";
```

Thoralf-M: This error is only used in the auction start and there the min bid is the price of the domain, not 1 IOTA

DaughterOfMars: Well then, maybe just remove the parenthesis bit?

DaughterOfMars (approved):

thibault-martinez (approved):

PR #131, opened 2025-04-10
feat(scripts): admin register names if provided in init script

### LINKED ISSUE #83: Initial list of names
Is there already a way to register an initial list of names from the init script? If not, add one.

Thoralf-M:
Added functionality so the admin can register names if provided in the init script

If there is a `scripts/init/namesToRegister.csv` file it will create an extra transaction to register the names provided there and sends them to the admin address. Later they can be sent to the respective addresses using the `transfer-names.ts` script

I removed the eslint-plugin-header dependency and check as it's limited to only a single header, we can't have
```
// Copyright (c) Mysten Labs, Inc.
```
and
```
// Copyright (c) 2025 IOTA Stiftung
```
with it

Fixes https://github.com/iotaledger/iota-names/issues/83

Tested by running `ts-node init/init.ts localnet 0x88031651332a044cca8261bbb12396a629e9f7d9fa9bcac70faae2f05c6e5d2f`
The names get registered and then since another admin address is provided also sent to the admin address
```
Registering 2 names: {
  '000.iota': '0x0000000000000000000000000000000000000000000000000000000000000000',
  'bitcoin.iota': undefined
}
```

Note: if one would try to register more than ~2040 names it would fail because of transaction limits and we don't split them into multiple txs


.eslintrc.cjs
```diff
@@ -54,15 +54,6 @@ module.exports = {
...
 		],
-		'header/header': [
-			2,
-			'line',
-			[
-				' Copyright (c) Mysten Labs, Inc.',
-				' Modifications Copyright (c) 2025 IOTA Stiftung',
-				' SPDX-License-Identifier: Apache-2.0',
-			],
-		],
```
thibault-martinez, line 65: How come?

scripts/init/setup.ts
```diff
@@ -46,6 +46,14 @@ export const setup = async (packageInfo: PackageInfo, network: string) => {
...
 	const txb = new Transaction();
 
+	// authorize admin module
+	authorizeApp({
+		txb,
+		adminCap: packageInfo.IotaNames.adminCap,
+		iotaNames: packageInfo.IotaNames.iotaNames,
+		type: `${packageInfo.IotaNames.packageId}::admin::Admin`,
+		iotaNamesPackageId: packageInfo.IotaNames.packageId,
+	});
```
thibault-martinez, line 56: Does it not fall under the for loop below?

.eslintrc.cjs
thibault-martinez: Ah, you explained, sorry

scripts/init/setup.ts
Thoralf-M: One could somehow merge it, but I didn't as this is just a module here and not an own package

scripts/reserved-names/reserve-names.ts
```diff
@@ -0,0 +1,51 @@
...
+	iotaNamesPackageId: string,
+	adminCap: string,
+	iotaNamesObjectId: string,
+) => {
+	return txb.moveCall({
+		target: `${iotaNamesPackageId}::admin::reserve_domains`,
+		arguments: [
+			txb.object(adminCap),
+			txb.object(iotaNamesObjectId),
+			txb.pure.vector('string', Object.keys(nameAddressPairs)),
```
thibault-martinez: So for my understanding, we disregard the addresses because `reserve_domains` send to ctx which is the admin? `transfer-names.ts` will later operate on the same `namesToRegister.csv` file to actually send them to the addresses?

Thoralf-M: The path for the files would be adjusted so `transfer-names.ts` actually makes use of it, but otherwise that's the idea, yes

scripts/init/names-to-register.csv
```diff
@@ -0,0 +1,4 @@
+# Names to register during the init process
+# Format (name without ".iota"): <name>,[address]
```
thibault-martinez:
```suggestion
# Format (name without ".iota"): <name>[,address]
```

scripts/init/names-to-register.csv
```diff
@@ -0,0 +1,4 @@
...
+000,0x0000000000000000000000000000000000000000000000000000000000000000
+bitcoin
```
thibault-martinez: I think I would actually comment them

Thoralf-M: https://github.com/iotaledger/iota-names/pull/131/commits/75deb1b0d6271e52495f51ae8128326978be7197

scripts/init/names-to-register.csv
Thoralf-M: https://github.com/iotaledger/iota-names/pull/131/commits/75deb1b0d6271e52495f51ae8128326978be7197

thibault-martinez: No?

Thoralf-M: Right, here https://github.com/iotaledger/iota-names/pull/131/commits/4198e5140ebe5e4f2fc4e363004e0d29c82dadde

scripts/reserved-names/reserve-names.ts
```diff
@@ -0,0 +1,59 @@
...
+
+import fs from 'fs';
+import { Transaction } from '@iota/iota-sdk/transactions';
+import { IOTA_CLOCK_OBJECT_ID, isValidIotaAddress } from '@iota/iota-sdk/utils';
+
+import { isValidIotaName, normalizeIotaName } from '../../sdk/src/utils';
+
+const YEARS_TO_RESERVE = 1;
+
+// Parses a CSV file with <name>,[address] pairs, names without ".iota"
```
thibault-martinez:
```suggestion
// Parses a CSV file with <name>[,address] pairs, names without ".iota"
```

Thoralf-M: Btw I made the code so it works also if there is a `,` without an address :trollface:

scripts/reserved-names/reserve-names.ts
```diff
@@ -0,0 +1,59 @@
...
+			}
+			nameAddressPairs[normalizedName] = address || undefined;
+			return undefined;
+		});
+	return nameAddressPairs;
+};
+
+export const reserveDomains = (
+	txb: Transaction,
+	nameAddressPairs: Record<string, string | undefined>,
```
thibault-martinez: Ok I'm annoying now so feel free to discard: I would just take the names, you already have `Object.keys` in init to check the length, just pass this, cleaner AP IMO

thibault-martinez (approved):

scripts/init/init.ts
```diff
@@ -35,12 +37,35 @@ export const init = async (
 	console.log('Published:', published);
 	await setup(published, network);
 
+	const client = getClient(network);
+	const packageInfo = readPackageInfo(network);
+
+	let namesToReserveFile = './init/names-to-register.csv';
+	if (fs.existsSync(namesToReserveFile)) {
+		const nameAddressPairs = parseCsvFile(namesToReserveFile);
```
DaughterOfMars, line 45: CSV seems like a weird format for KV pairs. Maybe json would be better?

thibault-martinez: I don't mind whether it's CSV or JSON but I wouldn't say CSV is weird, we already use it extensively for pairs in a lot of projects

DaughterOfMars: I guess that's fair. I don't use CSV a lot

scripts/reserved-names/reserve-names.ts
Thoralf-M: Makes sense, changed in https://github.com/iotaledger/iota-names/pull/131/commits/e5c64b95b445365df36583b92fb1b1a5f6fb2363

DaughterOfMars (approved):

PR #134, opened 2025-04-14
chore: Clean up name validation regex

### LINKED ISSUE #60: Double check utils.ts regexes
what to heck???? there's no way this regex is necessary 😆 obviously no need to fix it here but we should remember to look at this later

_Originally posted by @DaughterOfMars in https://github.com/iotaledger/iota-names/pull/23#discussion_r1963669873_

DaughterOfMars:
## Description

The `utils.ts` regexes are over-complicated. This PR simplifies these and nests them to ensure that there are fewer possibilities for errors.

Closes #60


Thoralf-M (changes requested):
```
console.log(isValidIotaName('invalidName.iota'))
console.log(isValidIotaName('invalid-.iota'))
```
both print true, but should be false

DaughterOfMars:
> ```
> console.log(isValidIotaName('invalidName.iota'))
> console.log(isValidIotaName('invalid-.iota'))
> ```
> 
> both print true, but should be false

The first one is valid, but I'll fix the second one

Thoralf-M:
In the smart contract uppercase is not allowed

DaughterOfMars:
> In the smart contract uppercase is not allowed

okay, I made them case sensitive then

Thoralf-M (changes requested):
Regex for .iota looks fine now, but this one returns false
`console.log(isValidIotaName('@name'))`

sdk/test/name-validation.ts
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
```
Thoralf-M: The eslint-plugin-header is limited to a single header for all files, I noticed the same problem in https://github.com/iotaledger/iota-names/pull/131 and removed it there, not sure if we want to do that

DaughterOfMars: Yeah that's pretty annoying, but I'd prefer just removing it over having these errors for now

sdk/test/name-validation.ts
```diff
@@ -0,0 +1,99 @@
...
+
+	test('should normalize as dot-style', () => {
+		expect(normalizeIotaName('@test', 'dot')).toEqual('test.iota');
+		expect(normalizeIotaName('test.iota', 'dot')).toEqual('test.iota');
+		expect(normalizeIotaName('sub.test@test', 'dot')).toEqual('sub.test.iota');
+		expect(normalizeIotaName('more.sub.test@test', 'dot')).toEqual('more.sub.test.iota');
+	});
+
+	test('should normalize as at-style', () => {
+		expect(normalizeIotaName('@test', 'dot')).toEqual('@test');
```
Thoralf-M: Should this end up in `test.iota`?

sdk/test/name-validation.ts
```diff
@@ -0,0 +1,99 @@
...
+		expect(normalizeIotaName('test@test', 'at')).toEqual('test@test');
+		expect(normalizeIotaName('sub.test.iota', 'at')).toEqual('sub.test@test');
+		expect(normalizeIotaName('more.sub.test.iota', 'at')).toEqual('more.sub.test@test');
+	});
+
+	test('should be invalid', () => {
```
Thoralf-M: Can you also add a test with capital letter and empty space/nothing please?
So for example `Name.iota` ` .iota` `space. .iota` `empty..iota` `.iota`

sdk/test/name-validation.ts
DaughterOfMars: oop, yeah

sdk/test/name-validation.ts
DaughterOfMars: done

Thoralf-M: Thanks

Thoralf-M (approved):

sdk/src/utils.ts
```diff
@@ -2,43 +2,36 @@
 // Modifications Copyright (c) 2025 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-const IOTA_NAME_REGEX =
-	/^(?!.*(^(?!@)|[-.@])($|[-.@]))(?:[a-z0-9-]{0,63}(?:\.[a-z0-9-]{0,63})*)?@[a-z0-9-]{0,63}$/i;
-const IOTA_NAMES_DOMAIN_REGEX = /^(?!.*(^|[-.])($|[-.]))(?:[a-z0-9-]{0,63}\.)+iota$/i;
-const MAX_IOTA_NAME_LENGTH = 235;
+const LABEL_REGEX = /(?!-)[a-z0-9-]{0,62}[a-z0-9]/;
```
thibault-martinez, line 5: Does `(?!-)` mean the first one can't be `-`?

DaughterOfMars: yep, exactly

sdk/src/utils.ts
```diff
@@ -2,43 +2,36 @@
...
+const PATH_REGEX = new RegExp(`(?:${LABEL_REGEX.source}(?:\\.${LABEL_REGEX.source})*)`);
+const NAME_REGEX = new RegExp(`^(${PATH_REGEX.source})?@${LABEL_REGEX.source}$`);
+const DOMAIN_REGEX = new RegExp(`^(?:${LABEL_REGEX.source}\\.)+(iota)$`);
```
thibault-martinez, line 8: Can't reuse `PATH_REGEX`  here? Could remove the `+` and move the `.` to `iota`. Not sure if worth it or not.

DaughterOfMars: could probably yeah

DaughterOfMars: but this regex is actually simpler

thibault-martinez: It has one more char, I wouldn't say it's simpler, but alright

thibault-martinez (approved):

PR #137, opened 2025-04-14
chore: Add move format check CI

### LINKED ISSUE #40: Move fmt CI check

DaughterOfMars:
## Description

Adds the `move_lint.yaml` CI workflow with a format check using the prettier plugin from Mystenlabs.

Closes #40


thibault-martinez:
is it configurable? Do we have a configuration file in the iota repo?

Thoralf-M:
> is it configurable? Do we have a configuration file in the iota repo?

Yes it is, here is one being added https://github.com/iotaledger/iota/pull/6385/files

thibault-martinez:
> Yes it is, here is one being added https://github.com/iotaledger/iota/pull/6385/files

Thanks, was asking because the indent seems pretty big in this PR

.github/workflows/move_lint.yaml
```diff
@@ -0,0 +1,26 @@
...
+  check-format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Install prettier plugin
+        run: npm i -g prettier @mysten/prettier-plugin-move
+
+      - name: Run format check
+        run: prettier-move -l packages/**
```
Thoralf-M: `-l` won't make the CI fail if something is wrong, no?

DaughterOfMars: I don't actually know for sure but I changed it because you're probably right

Thoralf-M: I'm wondering what is even considered an error, for example this mess just ends up in a warning
```move
public   struct 
IotaNames has key 
                                         {
    id:    UID,
}
```

```
[warn] packages/iota-names/sources/iota_names.move
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
```

DaughterOfMars: I believe it is only warnings, but it still fails the CI properly it seems like

Thoralf-M: Hm yeah looks like it

Thoralf-M (approved):

packages/.prettierrc
thibault-martinez, line 1: How come you didn't take the same file as #6385 ?

DaughterOfMars: Didn't know about it, and I like this better tbh

thibault-martinez: Ok but we really shouldn't have different Move format for different repositories...

DaughterOfMars: I'll create a PR to change it in the core then 😆

thibault-martinez: Feel free to open a PR but in the meantime we should apply their format because we have no guarantee they are going to accept it

DaughterOfMars: well that PR isn't even merge yet. I added a comment to change theirs

thibault-martinez: Thanks

thibault-martinez:
conflicts

packages/.prettierrc
thibault-martinez, line 1: If we are relying on defaults, any reason you picked some and not the others? I would either put none or all, this is a bit weird

DaughterOfMars: You're right, we don't need `useModuleLabel`

thibault-martinez: I'm confused now, these are the defaults
```
VSCode configuration for the extension is placed under prettierMove, these are the default settings of the extension:

{
    "prettierMove.tabWidth": 4,
    "prettierMove.printWidth": 100,
    "prettierMove.useModuleLabel": true,
    "prettierMove.autoGroupImports": "module",
    "prettierMove.enableErrorDebug": false,
    "prettierMove.wrapComments": false
}
```

DaughterOfMars: No, they aren't. At least not for the `prettier-move` command.

```
Format options:

  --arrow-parens <always|avoid>
                           Include parentheses around a sole arrow function parameter.
                           Defaults to always.
  --auto-group-imports <package|module>
                           Group all use imports by 'package', 'module' or 'none'.
                           Defaults to package.
  --bracket-same-line      Put > of opening tags on the last line instead of on a new line.
                           Defaults to false.
  --no-bracket-spacing     Do not print spaces between brackets.
  --embedded-language-formatting <auto|off>
                           Control how Prettier formats quoted code embedded in the file.
                           Defaults to auto.
  --enable-error-debug     Print ERROR nodes instead of throwing an error.
                           Defaults to false.
  --end-of-line <lf|crlf|cr|auto>
                           Which end of line characters to apply.
                           Defaults to lf.
  --experimental-operator-position <start|end>
                           Where to print operators when binary expressions wrap lines.
                           Defaults to end.
  --no-experimental-ternaries
                           Default behavior of ternaries; keep question marks on the same line as the consequent.
  --html-whitespace-sensitivity <css|strict|ignore>
                           How to handle whitespaces in HTML.
                           Defaults to css.
  --jsx-single-quote       Use single quotes in JSX.
                           Defaults to false.
  --object-wrap <preserve|collapse>
                           How to wrap object literals.
                           Defaults to preserve.
  --parser <flow|babel|babel-flow|babel-ts|typescript|acorn|espree|meriyah|css|less|scss|json|json5|jsonc|json-stringify|graphql|markdown|mdx|vue|yaml|glimmer|html|angular|lwc|move>
                           Which parser to use.
  --print-width <int>      The line length where Prettier will try wrap.
                           Defaults to 80.
  --prose-wrap <always|never|preserve>
                           How to wrap prose.
                           Defaults to preserve.
  --quote-props <as-needed|consistent|preserve>
                           Change when properties in objects are quoted.
                           Defaults to as-needed.
  --no-semi                Do not print semicolons, except at the beginning of lines which may need them.
  --single-attribute-per-line
                           Enforce single attribute per line in HTML, Vue and JSX.
                           Defaults to false.
  --single-quote           Use single quotes instead of double quotes.
                           Defaults to false.
  --tab-width <int>        Number of spaces per indentation level.
                           Defaults to 2.
  --trailing-comma <all|es5|none>
                           Print trailing commas wherever possible when multi-line.
                           Defaults to all.
  --use-module-label       Enable module labels instead of module with braces. This option will be ignored if there is more than one module in the file.
                           Defaults to true.
  --use-tabs               Indent with tabs instead of spaces.
                           Defaults to false.
  --vue-indent-script-and-style
                           Indent script and style tags in Vue files.
                           Defaults to false.
  --wrap-comments          Wrap comments to the next line if the line is too long.
                           Defaults to false.
```

DaughterOfMars: apparently the default `useModuleLabel` listed there is wrong 😆

thibault-martinez: I really think we'd be better of being explicit with everything to avoid default discrepancies between docs, extension and binary...

DaughterOfMars: So where do we draw the line? Should I put every single option listed above?

thibault-martinez: That's what I would personally do yeah, doesn't hurt and we're sure of the result

DaughterOfMars: Okay, I did that

.github/workflows/move_lint.yaml
```diff
@@ -0,0 +1,26 @@
...
+  check-format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Install prettier plugin
+        run: npm i -g prettier @mysten/prettier-plugin-move
+
+      - name: Run format check
+        run: prettier-move -c packages/**
```
thibault-martinez, line 26: Just making sure: this is really using `packages/.prettierrc`? No need to pass it explicitly?

DaughterOfMars: yep, it searches for it

.github/workflows/move_lint.yaml
```diff
@@ -0,0 +1,26 @@
+name: Lint move files
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/move_test.yaml"
```
thibault-martinez: Copy/paste error?

DaughterOfMars: oop, yeah

packages/.prettierrc
```diff
@@ -0,0 +1,26 @@
+{
+	"arrowParens": "always",
+	"autoGroupImports": "module",
+	"bracketSameLine": false,
+	"bracketSpacing": true,
+	"embeddedLanguageFormatting": "auto",
+	"enableErrorDebug": false,
+	"endOfLine": "lf",
+	"experimentalOperatorPosition": "end",
+	"experimentalTernaries": false,
```
thibault-martinez: `bracketSpacing` and `semi ` have the same pattern being a `--no-*` option but they are both true and this one is false, weird

DaughterOfMars: 🤷‍♀️

thibault-martinez: Where did you get this `false` from?

DaughterOfMars: double checked and I don't know why I put false, possibly it is the prettier suggested default

thibault-martinez (approved):

PR #138, opened 2025-04-14
feat(workflows): simplify client setup

Thoralf-M:
https://github.com/iotaledger/iota/issues/6203
Tested here https://github.com/iotaledger/iota-names/actions/runs/14446095586/job/40506941650?pr=138

### DIFF
--- a/.github/workflows/iota-names-build-tx.yaml
+++ b/.github/workflows/iota-names-build-tx.yaml
@@ -57,7 +57,6 @@ jobs:
       - name: YAML Setup
         run: |
           iota client --yes new-env --rpc ${{ inputs.rpc }} --alias network
-          iota client switch --env network
 
       - name: NPM BUILD TX Environment
         uses: actions/setup-node@cdca7365b2dadb8aad0a33bc7601856ffabcc48e # v4.3.0


.github/workflows/iota-names-build-tx.yaml
```diff
@@ -57,7 +57,6 @@ jobs:
       - name: YAML Setup
         run: |
           iota client --yes new-env --rpc ${{ inputs.rpc }} --alias network
-          iota client switch --env network
```
thibault-martinez, line 60: So how come we had this issue then? You added this recently and https://github.com/iotaledger/iota/pull/5330 s a bit old, was it just not released yet?

.github/workflows/iota-names-build-tx.yaml
```diff
@@ -57,7 +57,6 @@ jobs:
       - name: YAML Setup
         run: |
           iota client --yes new-env --rpc ${{ inputs.rpc }} --alias network
```
thibault-martinez, line 59: Do we even need the alias then?

.github/workflows/iota-names-build-tx.yaml
Thoralf-M: Yes I think that was the reason

.github/workflows/iota-names-build-tx.yaml
Thoralf-M: Yes, alias is required

thibault-martinez (approved):

DaughterOfMars (approved):

PR #177, opened 2025-05-02
feat(scripts)!: update transfer-names

### LINKED ISSUE #140: Adjust and test `transfer-names.ts`
The path for the files would be adjusted so `transfer-names.ts` actually makes use of it, but otherwise that's the idea, yes

_Originally posted by @Thoralf-M in https://github.com/iotaledger/iota-names/pull/131#discussion_r2044347655_

Thoralf-M:
Updated transfer-names script to work with any network, not just hardcoded mainnet

Also removed the GAS_OBJECT env variable one had to set manually, gas will be selected automatically now

Fixes #140 

Tested in a local network by modifying `scripts/names-to-register.csv` before running the init script and then updated `reserved-names/sample/sample.csv` to contain the same name, finally `ts-node reserved-names/transfer-names.ts`


marc2332:
Conflicts @Thoralf-M

thibault-martinez:
Let's solve conflicts and merge this

scripts/transactions/main-package-upgrade.ts
```diff
@@ -14,10 +15,15 @@ const network = process.env.NETWORK || 'mainnet';
 // Github actions are always on mainnet.
 const mainPackageUpgrade = async () => {
     const packageInfo = readPackageInfo(network);
-    const gasObjectId = process.env.GAS_OBJECT;
+    let client = getClient(network);
+    let gasCoinsPage = await client.getCoins({ owner: packageInfo.adminAddress });
+    const gasObjectId = gasCoinsPage.data.find(
+        (coin) => BigInt(coin.balance) >= 2000000000,
```
DaughterOfMars: where dis come from

Thoralf-M: I added this, so one doesn't has to manually provide the gas coin

DaughterOfMars: I mean the number, sorry

Thoralf-M: Ah, tbh I wasn't sure what to do, I uses this number as it should be enough to cover the gas fees for any operation, just using the first gas coin could fail if it has not enough amount
But maybe it would be better to just sort by amount and use the one with the highest amount, wdyt?

DaughterOfMars: I don't know 😭 somehow both seem wrong but I guess I'd prefer getting the biggest one

Thoralf-M: Well what else would you do? I changed it to highest balance here https://github.com/iotaledger/iota-names/pull/177/commits/fc523e2c6c7f18e7c9fb68c7becd30deb8640f11

DaughterOfMars (approved):

.github/workflows/iota-names-build-tx.yaml
```diff
@@ -4,6 +4,13 @@ name: Build IOTA-Names TX
 on:
   workflow_dispatch:
     inputs:
+      network:
+        description: "select the network to use"
+        type: choice
+        options:
+          - Mainnet
+          - Testnet
+          - Devnet
```
thibault-martinez: Does the case not matter?

scripts/reserved-names/fetch-owned-names.ts
```diff
@@ -8,9 +8,15 @@ import { IotaObjectResponse } from '@iota/iota-sdk/client';
 import { readPackageInfo } from '../package-info/constants';
 import { getClient } from '../utils/utils';
 
-const getAllOwnedDomains = async () => {
-    const packageInfo = readPackageInfo('mainnet');
-    let client = getClient('mainnet');
+export const getAllOwnedDomains = async () => {
```
thibault-martinez: Just wondering if you want to call this `getAllOwnedNames` for consistency

scripts/transactions/main-package-upgrade.ts
```diff
@@ -14,10 +15,15 @@ const network = process.env.NETWORK || 'mainnet';
 // Github actions are always on mainnet.
 const mainPackageUpgrade = async () => {
     const packageInfo = readPackageInfo(network);
-    const gasObjectId = process.env.GAS_OBJECT;
+    let client = getClient(network);
+    let gasCoinsPage = await client.getCoins({ owner: packageInfo.adminAddress });
+    const gasObjectId = gasCoinsPage.data.sort((a, b) =>
+        Number(BigInt(b.balance) - BigInt(a.balance)),
```
thibault-martinez, line 20: This is so weird, does the sort require a number and not an order/predicate ?

scripts/transactions/main-package-upgrade.ts
```diff
@@ -14,10 +15,15 @@ const network = process.env.NETWORK || 'mainnet';
...
+    )[0]?.coinObjectId;
 
     // Enabling the gas Object check only on mainnet, to allow testnet multisig tests.
-    if (!gasObjectId) throw new Error('No gas object supplied for a mainnet transaction');
+    if (!gasObjectId)
+        throw new Error('No gas object with sufficient balance found for a mainnet transaction');
```
thibault-martinez: We don't really know that it's insufficient?

scripts/utils/utils.ts
```diff
@@ -210,20 +210,14 @@ export const signAndExecute = async (txb: Transaction, network: string) => {
 export const prepareMultisigTx = async (tx: Transaction, network: string, address?: string) => {
     const adminAddress = address ?? getActiveAddress();
     const client = getClient(network);
-    const gasObjectId = process.env.GAS_OBJECT;
```
thibault-martinez, line 213: How is it picked up in this function then?

scripts/reserved-names/fetch-owned-names.ts
thibault-martinez: or `fetchOwnedNames`

scripts/transactions/main-package-upgrade.ts
Thoralf-M: Yes it returns a number:
```
- A negative value indicates that a should come before b.
- A positive value indicates that a should come after b.
- Zero or NaN indicates that a and b are considered equal.
```

scripts/transactions/main-package-upgrade.ts
Thoralf-M: True, not anymore since I changed it, will adapt the message

scripts/utils/utils.ts
Thoralf-M: We just get it from the client now
```
    let gasCoinsPage = await client.getCoins({ owner: packageInfo.adminAddress });
    const gasObjectId = gasCoinsPage.data.sort((a, b) =>
        Number(BigInt(b.balance) - BigInt(a.balance)),
    )[0]?.coinObjectId;
```

.github/workflows/iota-names-build-tx.yaml
Thoralf-M: It does, good catch

scripts/transactions/main-package-upgrade.ts
Thoralf-M: https://github.com/iotaledger/iota-names/pull/177/commits/af115cd9b15dcea617432e44013d349d9a50f07c

scripts/reserved-names/fetch-owned-names.ts
Thoralf-M: Renamed in https://github.com/iotaledger/iota-names/pull/177/commits/af115cd9b15dcea617432e44013d349d9a50f07c

.github/workflows/iota-names-build-tx.yaml
Thoralf-M: https://github.com/iotaledger/iota-names/pull/177/commits/af115cd9b15dcea617432e44013d349d9a50f07c

.github/workflows/iota-names-build-tx.yaml
```diff
@@ -75,8 +77,7 @@ jobs:
         if: ${{ inputs.transaction_type == 'Profits to Treasury' }}
         env:
           NODE_ENV: production
-          GAS_OBJECT: ${{ inputs.gas_object_id }}
-          NETWORK: mainnet
```
thibault-martinez, line 79: There are 2 more jobs that still have an hardcoded mainnet, are we changing them later?

scripts/utils/utils.ts
thibault-martinez: It's explicit in `scripts/transactions/main-package-upgrade.ts` because we're using the CLI and not here?

.github/workflows/iota-names-build-tx.yaml
Thoralf-M: Tbh not sure what the mainnet-setup.ts is for and for the ugprade I'm wondering how they do it in the testnet, if they don't use this script there (maybe just manually, because they don't have a multisig in the testnet, didn't check)
So depends on how we want to set our deployments up if we change them or not

scripts/utils/utils.ts
Thoralf-M: Yes, the TS SDK automatically selects gas coins from the sender address if you don't provide it

scripts/transactions/main-package-upgrade.ts
thibault-martinez: Ewww

scripts/transactions/main-package-upgrade.ts
```diff
@@ -6,18 +6,21 @@ import { execSync } from 'child_process';
...
 // if upgradeCap & gasObject is on mainnet, it has to be on mainnet.
-// Github actions are always on mainnet.
 const mainPackageUpgrade = async () => {
     const packageInfo = readPackageInfo(network);
-    const gasObjectId = process.env.GAS_OBJECT;
+    let client = getClient(network);
+    let gasCoinsPage = await client.getCoins({ owner: packageInfo.adminAddress });
+    const gasObjectId = gasCoinsPage.data.sort((a, b) =>
+        Number(BigInt(b.balance) - BigInt(a.balance)),
+    )[0]?.coinObjectId;
```
thibault-martinez, line 21: Just wondering if there is no function to get a gas coin from the client instead of repeating this

scripts/transactions/main-package-upgrade.ts
```diff
@@ -6,18 +6,21 @@ import { execSync } from 'child_process';
...
 const network = process.env.NETWORK || 'mainnet';
 
 // Active env of iota has to be the same with the env we're publishing to.
 // if upgradeCap & gasObject is on mainnet, it has to be on mainnet.
-// Github actions are always on mainnet.
 const mainPackageUpgrade = async () => {
     const packageInfo = readPackageInfo(network);
-    const gasObjectId = process.env.GAS_OBJECT;
+    let client = getClient(network);
+    let gasCoinsPage = await client.getCoins({ owner: packageInfo.adminAddress });
```
thibault-martinez, line 18: Just confirming, that would return the biggest amount of the first page only?

.github/workflows/iota-names-build-tx.yaml
thibault-martinez: Can you open an issue to check that later please?

scripts/reserved-names/transfer-names.ts
```diff
@@ -9,6 +9,21 @@ import { isValidIotaAddress } from '@iota/iota-sdk/utils';
...
 import { readPackageInfo } from '../package-info/constants';
 import { prepareMultisigTx } from '../utils/utils';
+import { fetchOwnedNames } from './fetch-owned-names';
+
+async function main() {
+    await fetchOwnedNames();
+
+    parseOwnedNamesObjects();
+
+    // Parses the `transfers.csv` file, and creates the list of object transfers
```
thibault-martinez, line 19: Is the comment correct? `transfers.csv`?

thibault-martinez (approved):

scripts/transactions/main-package-upgrade.ts
Thoralf-M: Yes, but that should be fine unless someone really wants to spam this address with dust (and then we can just merge them before calling this)

scripts/transactions/main-package-upgrade.ts
Thoralf-M: No, there is `setGasPayment()` which just set all available coins to the transaction builder in TS, but it doesn't return a single id

scripts/reserved-names/transfer-names.ts
Thoralf-M: Well the provided file needs to be changed before running it, added a comment

.github/workflows/iota-names-build-tx.yaml
Thoralf-M: Sure:
https://github.com/iotaledger/iota-names/issues/195
https://github.com/iotaledger/iota-names/issues/196

PR #186, opened 2025-05-13
chore: Add `run:example` script and fix the SDK

marc2332:


evavirseda (approved):

sdk/examples/example.ts
```diff
@@ -2,61 +2,86 @@
...
 import { Transaction } from '@iota/iota-sdk/transactions';
 
 import { IotaNamesClient } from '../src/iota-names-client.js';
 import { IotaNamesTransaction } from '../src/iota-names-transaction.js';
 
-// Initialize and execute the IotaNamesClient to fetch the renewal price list
 (async () => {
-    const network = 'testnet';
-    // Step 1: Create a IotaClient instance
+    const domain = `cool-domain-${(Math.random() * 10000000).toString().split('.').join('')}.iota`;
```
DaughterOfMars:
```suggestion
    const domain = `cool-domain-${Math.floor(Math.random() * 10000000)}.iota`;
```

Not actually sure why the number is there but this is maybe slightly better.

marc2332: Lol, true, the number is kinda useless in my code, I didnt think it much tbh

msarcev (approved):

PR #194, opened 2025-05-21
feat(iota-names): Split user data and metadata fields

DaughterOfMars:
## Description

Currently user data is used more like metadata, with predefined keys which are inflexible. This is counterintuitive because `user` implies that it can be set according to the user's wishes. 

This PR splits this field into two: One which has very limited access patterns for setting predefined metadata keys, and the other which allows anything.


thibault-martinez:
Not moving forward with this.

PR #203, opened 2025-05-22
feat: add registry events

thibault-martinez:


thibault-martinez:
Need to add one for leaves

packages/iota-names/sources/registry.move
```diff
@@ -221,7 +239,13 @@ public fun set_reverse_lookup(self: &mut Registry, address: address, domain: Dom
     assert!(target.is_some(), ETargetNotSet);
     assert!(some(address) == target, ERecordMismatch);
 
+    event::emit(IotaNamesReverseRegistryEvent {
+        key: address,
+        value: domain
+    });
+
     if (self.reverse_registry.contains(address)) {
+        // TODO different event for update and add?
```
Thoralf-M: If you don't see a need and want to add it now, I would remove this comment
We can just add later without this TODO, but otherwise it will be here for no reason

thibault-martinez: Just wanted to trigger the discussion, I'll remove it

Thoralf-M (approved):

DaughterOfMars (approved):

PR #218, opened 2025-05-28
feat(names-ts-sdk): Support custom networks in the iota names client

### LINKED ISSUE #209: Allow `IotaNamesClient` to be configured with custom networks
Extend the `IotaNamesClient` constructor (and possibility more) so that any network can be configured when creating an IOTA Names Client (eg localnet, alphanet, any custom network basically)

See for reference https://github.com/iotaledger/iota-names/blob/63911c2234d0c73ceb6a538cd304c13070bb30f5/sdk/src/iota-names-client.ts#L32

marc2332:
Closes https://github.com/iotaledger/iota-names/issues/209


sdk/test/pre-built.ts
marc2332, line 1: I commented this file because `mainnet` and `testnet` are not supported networks in iota names, so the type fails as they are not defined in the `packages` constant.

evavirseda (approved):

msarcev (approved):

PR #230, opened 2025-06-02
feat(indexer): add name length distribution metric

### LINKED ISSUE #7031: IOTA-Names metrics: domain length distribution

Thoralf-M:
Add a new metric to track the count of registered names with for the length of the second level name

Fixes https://github.com/iotaledger/iota/issues/7031

Tested in a localnet:
image

### DIFF
--- a/indexer/src/worker.rs
+++ b/indexer/src/worker.rs
@@ -10,7 +10,7 @@ use iota_data_ingestion_core::{
     DataIngestionMetrics, FileProgressStore, IndexerExecutor, ReaderOptions, Worker, WorkerPool,
 };
 use iota_json::IotaJsonValue;
-use iota_names::config::IotaNamesConfig;
+use iota_names::{config::IotaNamesConfig, domain::Domain};
 use iota_types::{
     Identifier, TypeTag,
     balance::Balance,
@@ -87,8 +87,14 @@ impl IotaNamesWorker {
 
     fn process_event(&self, event: IotaNamesEvent) -> anyhow::Result<()> {
         match event {
-            IotaNamesEvent::IotaNamesRegistry(_event) => {
+            IotaNamesEvent::IotaNamesRegistry(event) => {
                 self.metrics.total_name_records.inc();
+                let domain = Domain::from_str(&event.domain)?;
+                let second_level_name_len = domain.label(1).expect("missing SDL").len();
+                self.metrics
+                    .name_length_distribution
+                    .with_label_values(&[&second_level_name_len.to_string()])
+                    .inc();
             }
             IotaNamesEvent::IotaNamesReverseRegistry(_event) => (),
             IotaNamesEvent::AuctionStarted(_event) => {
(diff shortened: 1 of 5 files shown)


indexer/src/worker.rs
```diff
@@ -77,8 +77,14 @@ impl IotaNamesWorker {
 
     fn process_event(&self, event: IotaNamesEvent) -> anyhow::Result<()> {
         match event {
-            IotaNamesEvent::IotaNamesRegistry(_event) => {
+            IotaNamesEvent::IotaNamesRegistry(event) => {
                 self.metrics.total_name_records.add(1);
+                let domain = Domain::from_str(&event.domain)?;
+                let second_level_name_len = domain.label(1).expect("missing SDL").len();
```
thibault-martinez:
```suggestion
                let second_level_name_len = domain.label(1).expect("missing SLD").len();
```

thibault-martinez:
Feel free to do it in another PR but it would be nice to also add a box with the average

indexer/src/worker.rs
```diff
@@ -78,8 +78,14 @@ impl IotaNamesWorker {
 
     fn process_event(&self, event: IotaNamesEvent) -> anyhow::Result<()> {
         match event {
-            IotaNamesEvent::IotaNamesRegistry(_event) => {
+            IotaNamesEvent::IotaNamesRegistry(event) => {
                 self.metrics.total_name_records.add(1);
+                let domain = Domain::from_str(&event.domain)?;
+                let second_level_name_len = domain.label(1).expect("missing SLD").len();
```
DaughterOfMars: Isn't there a `.sld()` fn?

Thoralf-M: No, just `.is_sld()`

DaughterOfMars: sad

Thoralf-M:
> Feel free to do it in another PR but it would be nice to also add a box with the average

Do you mean an own view with the average just as number or what?

thibault-martinez:
> > Feel free to do it in another PR but it would be nice to also add a box with the average
> 
> Do you mean an own view with the average just as number

Yes

thibault-martinez:
conflicts

indexer/src/worker.rs
```diff
@@ -87,8 +87,14 @@ impl IotaNamesWorker {
 
     fn process_event(&self, event: IotaNamesEvent) -> anyhow::Result<()> {
         match event {
-            IotaNamesEvent::IotaNamesRegistry(_event) => {
+            IotaNamesEvent::IotaNamesRegistry(event) => {
                 self.metrics.total_name_records.inc();
+                let domain = Domain::from_str(&event.domain)?;
```
thibault-martinez: I'm seeing now that `IotaNamesRegistry` is the only event where the domain is a string and not a Domain, maybe we should change that? We actually convert it to a string in the SCs

Thoralf-M: Good point, changed here https://github.com/iotaledger/iota-names/pull/230/commits/9a8fdbf58735539bd06f0e346fd06b45896f9609

thibault-martinez (approved):

PR #234, opened 2025-06-02
fix(indexer): fix balance decoding

Thoralf-M:
Tested in a localnet
image

### DIFF
--- a/indexer/src/worker.rs
+++ b/indexer/src/worker.rs
@@ -13,6 +13,7 @@ use iota_names::config::IotaNamesConfig;
 use iota_types::{
     Identifier, TypeTag,
     balance::Balance,
+    dynamic_field::Field,
     effects::{TransactionEffects, TransactionEffectsAPI},
     execution_status::ExecutionStatus,
     full_checkpoint_content::CheckpointData,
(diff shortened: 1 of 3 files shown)


indexer/src/worker.rs
```diff
@@ -142,7 +143,16 @@ impl Worker for IotaNamesWorker {
 
         for object in checkpoint.latest_live_output_objects() {
             if object.id() == balance_object_id {
-                let balance: Balance = object.to_rust().expect("invalid balance object");
+                let balance = bcs::from_bytes::<Field<MoveTypeLayout, Balance>>(
```
thibault-martinez, line 146: nice

Alex6323 (approved):

indexer/src/worker.rs
thibault-martinez: I thought it would end up more complex than that

thibault-martinez (approved):

indexer/src/worker.rs
Thoralf-M: Once you know that you need it like this, yes :joy:

PR #257, opened 2025-06-04
feat(dapp): add protected `/my-names` route

### LINKED ISSUE #246: Add protected `/my-names` route
- Add route, no need to make any fetching logic yet
- Add protected `/my-names` route which should be visible only if the wallet is connected, otherwise redirect to homepage
- Display item in the navbar if route is allowed
- Probably best is implementing a connection guard or similar for protected routes

evavirseda:
fixes https://github.com/iotaledger/iota-names/issues/246

### DIFF
--- a/dapp/src/app/(protected)/layout.tsx
+++ b/dapp/src/app/(protected)/layout.tsx
@@ -7,20 +7,7 @@ import { useCurrentAccount } from '@iota/dapp-kit';
 import { redirect } from 'next/navigation';
 import { useEffect, type PropsWithChildren } from 'react';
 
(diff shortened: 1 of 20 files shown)


begonaalvarezd (changes requested):
I see when im connected it always forces me to go to `my-names`. The user should not be forced to go there, the homepage is where the user searches for a name, and then `my-names` is where the user will manage its names, is a route the user can decide to go or not.

msarcev (approved):

begonaalvarezd (approved):

PR #261, opened 2025-06-04
feat(dapp): Support setting a target address and reverse lookup

### LINKED ISSUE #251: Add action to make name default (reverse lookup)
Depends on https://github.com/iotaledger/iota/pull/7211
Depends on https://github.com/iotaledger/iota-names/issues/247 & https://github.com/iotaledger/iota-names/issues/248, though it can be worked on the logic to set reverse lookups and allowance criteria (plus edge cases)

- Making a name as default means that, given an address, you are configuring a unique domain for such address. This is called Reverse Lookup
- To all names and subnames, we should render a button to "make name default" if possible (use `set_reverse_lookup` providing the connected address)
- Making the name default is possible if
    - its not expired
    - its not yet the default one
- You can know if its already the default one by using the reverse lookup query
- Add logic to set the reverse lookup. When setting the reverse lookup, if the target address (lookup) is not set or it is different to the new reverse lookup value, the target address should also be updated (`set_target_address`)

### LINKED ISSUE #252: Add action to set target address (lookup)
Depends on https://github.com/iotaledger/iota/pull/7211

Depends on https://github.com/iotaledger/iota-names/issues/247 & https://github.com/iotaledger/iota-names/issues/248, though it can be worked on the logic to set lookups and allowance criteria

⚠ Probably needs alignment with https://github.com/iotaledger/iota-names/issues/251 to not overlap
 
- Setting a target address means that given a domain, you are configuring to what address it resolves. This is called Lookup
- To all names and subnames, we should render a button to set target address if possible (use `set_target_address` providing the connected address)
- Setting a target address is possible if its not expired
- Render the view as
    - if target address is not yet set, inform about it
    - render an input where the user can enter an IOTA address
    - validate the IOTA address
    - prefill the address is its already set
- Its not necessary to modify the reverse lookup since it seems to be handled internally by the smart contracts (read the end of the issue description of https://github.com/iotaledger/iota-names/issues/251 to understand this note) .

marc2332:
Closes https://github.com/iotaledger/iota-names/issues/252
Closes https://github.com/iotaledger/iota-names/issues/251


KeitoTadashi (approved):
LGTM :rocket:

msarcev (approved):

PR #262, opened 2025-06-04
feat(dapp): List top level names

### LINKED ISSUE #247: Fetch and display all top level names
Depends on https://github.com/iotaledger/iota-names/issues/246, though it can be worked on the logic to fetch

- Fetch and display (just create a list with its params) all top level names by looking at owned objects with type `{packageid}::iota_names_registration::IotaNamesRegistration`
- Place them under a "names" section
- Use hooks with transtack
- Type the response accordingly with that it can be returned

brancoder:
resolves #247

### DIFF
--- a/dapp/src/app/(protected)/my-names/page.tsx
+++ b/dapp/src/app/(protected)/my-names/page.tsx
@@ -9,7 +9,7 @@ import { AvailabilityCheck } from '@/components';
 
 function MyNamesPage(): JSX.Element {
     return (
-        <div className="flex flex-col w-full gap-y-4 items-center">
+        <div className="flex flex-col w-full gap-y-lg items-center">
             <AvailabilityCheck />
             <div className="pt-md">
                 <Title title="My names" testId="my-names-page" />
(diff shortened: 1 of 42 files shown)


marc2332 (approved):

dapp/src/app/(protected)/my-names/page.tsx
```diff
@@ -3,17 +3,30 @@
 
 'use client';
 
-import { Title } from '@iota/apps-ui-kit';
+import { KeyValueInfo, Title } from '@iota/apps-ui-kit';
 
 import { AvailabilityCheck } from '@/components';
+import { useRegistrationNfts } from '@/hooks/useRegistrationNfts';
```
evavirseda: can we create an index file in the hooks folder and use the path `@/hooks`, like we do with components?

evavirseda (changes requested):
just a tiny detail, but works great!

dapp/src/app/(protected)/my-names/page.tsx
marc2332: I just pushed but I dont think this is a good idea, it leads to circular dependencies, unnecessary work (to add, update and remove the barrel files), and worse performance (as the moment you import a hook you end up loading all of them)

Literally everything is downsides, all you gain is shorter imports

evavirseda (approved):

Issue #253, opened 2025-06-03
Add action to set permissions


begonaalvarezd:
Depends on https://github.com/iotaledger/iota-names/issues/247 & https://github.com/iotaledger/iota-names/issues/248, though it can be worked on the logic to set permissions and allowance criteria

- For subnames (not names) we should render a button to set permissions if possible
- Its possible to set permissions if not expired
- Pre-fill existing permissions, use checkboxes
- Permissions are:
    - Allow subdomains
    - Allow renewal
- Add logic to update permissions (ptb move call)

PR #266, opened 2025-06-05
feat(dapp): action to create subname

### LINKED ISSUE #249: Add action to add subname
Depends on https://github.com/iotaledger/iota-names/issues/247 & https://github.com/iotaledger/iota-names/issues/248, though it can be worked on the logic to create subnames and allowance criteria

- To all names and subnames, we should render a button to add a subname
- Add is allowed if:
    - not expired: ignore grace periods here, look at expiration value only
    - subnames must have subname creation allowed, names always can have subnames
- Implement logic, ptb move call
- Use the same expiration time as parent for now

KeitoTadashi:
fixes #249 
depends on #253 :warning:


dapp/src/components/dialogs/UpdateNameDialog.tsx
```diff
@@ -108,6 +185,37 @@ export function UpdateNameDialog({ name, open, setOpen }: UpdateNameDialogProps)
...
+        const parentObjectId = getSubdomainObjectId(nameRecord.nameRecord.name);
+        updates.push({
+            type: 'new-subdomain',
+            subdomainName: fullSubdomainName,
+            parentNftId: parentObjectId ?? '',
+            expirationTimeParent: nameRecord?.nameRecord?.expirationTimestampMs || 0,
+            allowChildCreation: editIsAllowSubnames,
+            allowTimeExtension: editIsAllowingRenew,
+        });
+    }
```
brancoder: this breaks the transaction creation for changing subdomain permissions. it trys to add a new domain when just changing permission of the existing subdomain

brancoder (changes requested):

PR #268, opened 2025-06-05
improve search validation

### LINKED ISSUE #245: Improve search name input validation
- the user should be able to search at any time to be able to know what is wrong with the validation
- `.iota` should not be part of the input, we should add it ourselves
- domains that do not meet the criteria should display an error message as part of the input (use the error message field provided in the input)
    - subdomains should not be allowed, so no "." in the searched names. That affects to the current required `.iota`, it should not be part of the input, we should add it ourselves
    - domains that do not follow the regex rules should explain why (eg Invalid characters in the name. Allowed characters are a-z, 0-9, and - as long as it doesn’t appear at the beginning or end of the string, we should rephrase this a bit so its not a 1:1 copy to suins)
    - names have a min and max size which we have to also provide to the user, right now we display "name not priced" but that means nothing to the user, based on the pricelist we can know which is the min and the max allowed (this is a dapp validation only), so we should probably get the pricelist and use that as a dapp frontend validation criteria

evavirseda:
fixes https://github.com/iotaledger/iota-names/issues/245


dapp/src/components/AvailabilityCheck.tsx
```diff
@@ -17,59 +16,89 @@ export function AvailabilityCheck() {
...
+    const maxLengthAllowed = priceList?.maxLength;
 
-    const handleOnSearchChange = (value: string) => {
-        if (name.length > 0) {
-            setName('');
+    function normalizeNameInput(name: string) {
+        return name.toLowerCase().replace(/\.iota$/i, '');
+    }
+
+    function getValidationError(name: string): string | null {
```
marc2332: Can you move these functions outside of the component, but keep them in the same file?

dapp/src/components/AvailabilityCheck.tsx
```diff
@@ -17,59 +16,89 @@ export function AvailabilityCheck() {
...
+            return 'Invalid characters. Only a-z, 0-9, and hyphens (not at the beginning or end) are allowed';
         }
-        setSearchValue(value);
-    };
 
-    const handleOnSearch = async () => {
-        setName(searchValue);
-    };
+        const minLength = minLengthAllowed ?? 3;
+        const maxLength = maxLengthAllowed ?? 64;
```
marc2332:
```suggestion
        const minLength = priceList?.minLength ?? 3;
        const maxLength = priceList?.maxLength ?? 3;
```

dapp/src/components/AvailabilityCheck.tsx
```diff
@@ -17,59 +16,89 @@ export function AvailabilityCheck() {
...
     const [name, setName] = useState<string>('');
+    const [isPurchaseDialogOpen, setPurchaseDialogOpen] = useState(false);
 
     const { data, error } = useNameRecord(name);
+    const { data: priceList } = usePriceList();
 
-    const isValid = useMemo(() => isValidIotaName(searchValue), [searchValue]);
-    const [isPurchaseDialogOpen, setPurchaseDialogOpen] = useState(false);
+    const minLengthAllowed = priceList?.minLength;
+    const maxLengthAllowed = priceList?.maxLength;
```
marc2332:
```suggestion
```

dapp/src/components/AvailabilityCheck.tsx
```diff
@@ -85,9 +114,9 @@ export function AvailabilityCheck() {
                 </div>
             )}
 
-            {data?.type == 'available' && (
+            {canBuy && (
                 <div className="flex items-center space-x-4">
-                    <div className="text-body-md">Price: {formatNanosToIota(data.price)}</div>
+                    <div className="text-body-md">Price: {formatNanosToIota(data!.price)}</div>
```
marc2332:
```suggestion
                    <div className="text-body-md">Price: {formatNanosToIota(data.price)}</div>
```

dapp/src/components/AvailabilityCheck.tsx
```diff
@@ -17,59 +16,89 @@ export function AvailabilityCheck() {
...
+                        <p className="text-primary-20 dark:text-primary-80 text-label-lg">@</p>
+                    }
                 />
+
                 <Button
                     size={ButtonSize.Medium}
                     text="Search"
-                    disabled={!enableSearch}
-                    onClick={handleOnSearch}
+                    disabled={!enableSearch || Boolean(errorMessage)}
```
marc2332:
```suggestion
                    disabled={!enableSearch || !!errorMessage}
```

dapp/src/components/AvailabilityCheck.tsx
```diff
@@ -17,59 +16,89 @@ export function AvailabilityCheck() {
...
     function handlePurchase() {
         setPurchaseDialogOpen(false);
         setSearchValue('');
         setName('');
     }
 
+    const validationError = useMemo(() => getValidationError(searchValue), [searchValue]);
+    const handleSearch = useCallback(() => {
+        if (searchValue) setName(`${searchValue}.iota`);
+    }, [searchValue]);
```
marc2332: Move these hook calls before all the `handleBlabla` functions (code consistency)

marc2332: Also leave an empty line between both

marc2332 (changes requested):

brancoder (approved):
LGTM

dapp/src/components/AvailabilityCheck.tsx
```diff
@@ -5,71 +5,103 @@
...
+                        <p className="text-primary-20 dark:text-primary-80 text-label-lg">@</p>
+                    }
                 />
+
                 <Button
                     size={ButtonSize.Medium}
                     text="Search"
-                    disabled={!enableSearch}
-                    onClick={handleOnSearch}
+                    disabled={!enableSearch || !!errorMessage}
```
marc2332: I would move `!!errorMessage` to the definition of `enableSearch` itself

marc2332 (approved):
rest lgtm

PR #270, opened 2025-06-05
feat(dapp): show subdomains

### LINKED ISSUE #248: Fetch and display all node subnames
Depends on https://github.com/iotaledger/iota-names/issues/246, though it can be worked on the logic to fetch
Align with https://github.com/iotaledger/iota-names/issues/247 to reuse components if needed

- We ignore leafs for now
- Fetch and display (just create a list with its params) all node subnames by looking at owned objects with type `{packageid}::subdomain_registration::SubdomainRegistration`
- Place them under a "subnames" section
- Use hooks with transtack
- Type the response accordingly with that it can be returned

panteleymonchuk:
closes #248


dapp/src/hooks/useSubdomainRegistrations.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import { useGetAllOwnedObjects } from './useGetAllOwnedObjects';
+
+interface RegistrationNft {
+    name: string;
+    description?: string;
+    image_url?: string;
+    link?: string;
+    project_url?: string;
+}
```
brancoder: we could extract this type to type because we have a same one in `useRegistrationNfts`

brancoder (changes requested):
LG just a type change

dapp/src/hooks/useSubdomainRegistrations.ts
brancoder, line 1: also this hook looks same as useRegistrationNfts, they just differ in the object type.  maybe we could use same hook and add prop type. wdyt?

brancoder (changes requested):

msarcev (dismissed):
I wouldn't merge this until we merge https://github.com/iotaledger/iota-names/pull/277
No point in merging using the jsonRPC transport and then later having to fix it again cause it doesn't work on graphqlTransport

brancoder:
> I wouldn't merge this until we merge #277 No point in merging using the jsonRPC transport and then later having to fix it again cause it doesn't work on graphqlTransport

I think we can merge this,  this will work because the underlying transport will just change and hook will remain the same.

marc2332 (dismissed):
~~I would use the graphql query instead~~

dapp/src/app/(protected)/my-names/page.tsx
```diff
@@ -4,12 +4,25 @@
...
-    const registrationNfts = useRegistrationNfts();
+    const { iotaNamesClient } = useIotaNamesClient();
+    const packageId = iotaNamesClient.config.packageId;
+
+    const registrationNfts = useRegistrationNfts({
+        StructType: getIotaNamesRegistrationType(packageId),
+    });
+    const subdomains = useRegistrationNfts({
+        StructType: getIotaSubdomainRegistrationType(packageId),
+    });
```
marc2332: nit: consistency
```suggestion
    const domainsNfts = useRegistrationNfts({
        type: 'domains'
    });
    const subdomainsNfts = useRegistrationNfts({
        type: 'subdomains',
    });
```

dapp/src/app/(protected)/my-names/page.tsx
```diff
@@ -19,14 +32,31 @@ function MyNamesPage(): JSX.Element {
             </div>
             <div className="flex flex-col gap-x-sm items-center">
                 {registrationNfts?.map((nameRecord) => (
-                    <KeyValueInfo
-                        key={nameRecord.name}
-                        keyText={nameRecord.name}
-                        value={nameRecord?.description ?? ''}
-                        fullwidth
-                    />
+                    <>
```
marc2332: This fragment can be removed

dapp/src/hooks/useRegistrationNfts.ts
marc2332, line 1: I would maybe have kept the code of the filter here and instead just made the hook accept a `type` flag with `'domains' | 'subdomains'` type, having to pollute the components with filters is unnecessary

marc2332 (changes requested):

brancoder (changes requested):
conflicts 🙏

evavirseda (approved):

marc2332 (approved):

brancoder (approved):
LGTM

PR #272, opened 2025-06-06
fix(docker): add cache bust, so it compiles the actual binary

Thoralf-M:
Tested with `DOCKER_BUILDKIT=1 docker compose -f indexer/docker/docker-compose.yml build --build-arg CACHEBUST=$(date +%s) --progress=plain`
The main.rs timestamp had to be updated with `touch src/main.rs`, otherwise cargo wouldn't notice that there was a change

### DIFF
--- a/indexer/docker/Dockerfile
+++ b/indexer/docker/Dockerfile
@@ -37,6 +37,8 @@ RUN rm -rf src
 # 5. Copy the rest of the source code
 COPY ./ .
 
+ARG CACHEBUST=1
+RUN echo "Cache bust: $CACHEBUST" && touch src/main.rs
 # 6. Build the actual project
 RUN cargo build --profile ${PROFILE} --features ${CARGO_BUILD_FEATURES:=default}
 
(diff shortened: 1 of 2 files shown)


indexer/docker/Dockerfile
```diff
@@ -37,6 +37,8 @@ RUN rm -rf src
 # 5. Copy the rest of the source code
 COPY ./ .
 
+ARG CACHEBUST=1
+RUN echo "Cache bust: $CACHEBUST" && touch src/main.rs
```
thibault-martinez, line 41: So it always displays 1 ?

thibault-martinez (approved):

indexer/docker/Dockerfile
Thoralf-M: No, it assigns the cachebust argument (`--build-arg CACHEBUST=$(date +%s)`) to `CACHEBUST`, 1 is just default fallback value

thibault-martinez: I see, thanks

Alex6323 (approved):
lgtm

PR #275, opened 2025-06-06
add graphql for registration nfts

### LINKED ISSUE #274: Add graphql for fetching top level names
switch to graphql for  fetching top level names

brancoder:
resolves #274


brancoder:
not needed because transport has changed to graphql  https://github.com/iotaledger/iota-names/pull/277

PR #276, opened 2025-06-06
feat(indexer): add processed checkpoints and restructure grafana dashboard

Thoralf-M:
Add processed checkpoints view and restructure grafana dashboard so it looks like this:
image


thibault-martinez:
While you're at it, do you think you could display the balance in IOTA and not NANO? Separate PR is fine otherwise.

thibault-martinez (approved):

thibault-martinez:
Why is processed checkpoints in red btw?

Thoralf-M:
> Why is processed checkpoints in red btw?

Was the default, I didn't set any color

Alex6323 (approved):

PR #281, opened 2025-06-10
chore(indexer): decrease name record metrics

thibault-martinez:
They are just being increased atm, never decreased.

image
image
image
image

### DIFF
--- a/indexer/src/worker.rs
+++ b/indexer/src/worker.rs
@@ -109,7 +109,7 @@ impl IotaNamesWorker {
     fn process_event(&self, event: IotaNamesEvent) -> anyhow::Result<()> {
         match event {
             IotaNamesEvent::NameRecordAdded(event) => {
-                self.metrics.total_name_records.inc();
+                self.metrics.total_name_records_added.inc();
                 let second_level_name_len = event.domain.label(1).expect("missing SLD").len();
                 self.metrics
                     .name_length_distribution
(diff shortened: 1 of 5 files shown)


indexer/src/worker.rs
```diff
@@ -108,15 +108,23 @@ impl IotaNamesWorker {
...
                 self.metrics.total_name_records.inc();
                 let second_level_name_len = event.domain.label(1).expect("missing SLD").len();
                 self.metrics
                     .name_length_distribution
                     .with_label_values(&[&second_level_name_len.to_string()])
                     .inc();
             }
-            IotaNamesEvent::IotaNamesReverseRegistry(_event) => (),
+            IotaNamesEvent::NameRecordRemoved(event) => {
+                self.metrics.total_name_records.dec();
```
Thoralf-M: If we decrease this one, then we should have another one that never decreases

thibault-martinez: Why? Do you really think it's interesting to have max values instead of current values?

Thoralf-M: Not for everything, but I think it's interesting to have a number for the number of registrations since start

thibault-martinez: I split added/removed metrics and added a new display for the added one

Thoralf-M (approved):

DaughterOfMars (approved):

PR #283, opened 2025-06-11
feat(indexer): add target/default address metrics

### LINKED ISSUE #7028: IOTA-Names metrics: number of lookup / reverse lookup fields set

thibault-martinez:
Fixes https://github.com/iotaledger/iota/issues/7028

image

image


Thoralf-M:
What about the grafana file?

thibault-martinez:
> What about the grafana file?

Committed but forgot to push it, done, thanks

Thoralf-M (approved):

indexer/src/metrics.rs
```diff
@@ -73,6 +75,18 @@ impl IotaNamesMetrics {
...
             .unwrap(),
+            total_target_address: register_int_gauge_with_registry!(
+                "total_target_address",
+                "The total number of target addresses set",
+                registry,
+            )
+            .unwrap(),
+            total_default_address: register_int_gauge_with_registry!(
+                "total_default_address",
+                "The total number of default addresses set",
```
DaughterOfMars: Isn't it more accurate to say this is the number of default names (for addresses)?

thibault-martinez: Because target address and reverse lookup need to be sync, default name and default address can be used interchangeably (and actually are in SuiNS) so I think it's fine. If you feel strongly about it, I can change it.

DaughterOfMars: it definitely feels wrong to me

thibault-martinez: I maintain that it's correct but I will change it for you

thibault-martinez: done

DaughterOfMars: And the other fields?

thibault-martinez: Which ones?

packages/iota-names/sources/registry.move
```diff
@@ -59,11 +59,21 @@ public struct NameRecordRemovedEvent has copy, drop {
...
+    target_address: Option<address>
+}
+
 public struct ReverseLookupSetEvent has copy, drop {
     default_address: address,
     domain: Domain
 }
 
+public struct ReverseLookupUnsetEvent has copy, drop {
+    default_address: address,
```
DaughterOfMars, line 73: This one for instance

thibault-martinez: We can't call it `address` though. Funnily enough, you are the one that proposed `default_address` for this field. I would keep them, it's fine tbh.

DaughterOfMars: Oh, that's right :c The other one should be `default_domain` then?

thibault-martinez: https://github.com/iotaledger/iota-names/pull/283/commits/8ac09b05df33f7eb381f8dc283b77925040ce4c3

Thoralf-M (approved):

DaughterOfMars (approved):

PR #288, opened 2025-06-12
feat(indexer): persist all data and restore after restart

### LINKED ISSUE #7096: iota-names-indexer: metric numbers should not be reset when the indexer is restarted
## Description

The metrics should be initialized with potential values store in prometheus
https://github.com/iotaledger/iota/blob/13364dce67507773dc4977ed8d1114b11d88c036/crates/iota-names-indexer/src/worker.rs#L36-L41

## Motivation

Currently the metrics are initialized with 0 every time the indexer is restarted and therefore the numbers are wrong after a restart, unless we start to sync from checkpoint 0.

Thoralf-M:
Persist all data and restore after restart

Fixes: https://github.com/iotaledger/iota/issues/7096

Tested by stopping and restarting, continues with the checkpoints as expected and also checked the grafana board manually, it also contains the old state and the metrics continue from there
```
iota-names-indexer  | 2025-06-12T10:14:35.697692Z DEBUG iota_names_indexer::worker: Processing checkpoint: 1042
iota-names-indexer  | 2025-06-12T10:14:36.107153Z DEBUG iota_names_indexer::worker: Processing checkpoint: 1043
Gracefully stopping... (press Ctrl+C again to force)
[+] Stopping 3/3
 ✔ Container iota-names-indexer  Stopped                                                                                                            0.2s 
 ✔ Container grafana             Stopped                                                                                                            0.2s 
 ✔ Container prometheus          Stopped                                                                                                            0.2s 
 ~/Desktop/iota-names   feat/indexer-auction-price-metric ±✚  docker compose -f indexer/docker/docker-compose.yml up
[+] Running 3/3
 ✔ Container grafana             Created                                                                                                            0.0s 
 ✔ Container prometheus          Created                                                                                                            0.0s 
 ✔ Container iota-names-indexer  Created 
iota-names-indexer  | 2025-06-12T10:14:47.429438Z  INFO iota_names_indexer: Starting IOTA Names Indexer
iota-names-indexer  | 2025-06-12T10:14:47.490133Z  INFO iota_names_indexer::metrics: Successfully restored metrics from Prometheus
iota-names-indexer  | 2025-06-12T10:14:47.490786Z  INFO iota_names_indexer: Successfully restored metrics from Prometheus
iota-names-indexer  | 2025-06-12T10:14:47.490881Z  INFO iota_names_indexer::metrics: Starting prometheus metrics at 0.0.0.0:9189
iota-names-indexer  | 2025-06-12T10:14:47.491801Z  INFO iota_names_indexer::worker: Connecting to node at http://host.docker.internal:9000
iota-names-indexer  | 2025-06-12T10:14:47.754086Z DEBUG iota_names_indexer::worker: Processing checkpoint: 1044
iota-names-indexer  | 2025-06-12T10:14:47.754214Z DEBUG iota_names_indexer::worker: Processing checkpoint: 1045
```

Before restart: 
image
After restart:
image


DaughterOfMars (approved):

indexer/src/worker.rs
```diff
@@ -68,7 +68,7 @@ pub(crate) async fn run_iota_names_reader(
         executor
             .run(
                 // path to a local directory where checkpoints are stored.
-                PathBuf::from("./chk"),
+                PathBuf::from("./data/chk"),
```
thibault-martinez, line 74: Is this even used?

indexer/docker/docker-compose.yml
```diff
@@ -39,6 +41,7 @@ services:
     command:
       - "start"
       - "--node-url=http://host.docker.internal:9000"
+      - "--prometheus-url=http://prometheus:9090"
```
thibault-martinez, line 44: Is it not supposed to be `http://localhost:9090`?

DaughterOfMars: It works because the name is assigned by docker I think

indexer/src/worker.rs
Thoralf-M: At the moment not

indexer/src/main.rs
```diff
@@ -62,11 +65,20 @@ impl Command {
...
             } => {
                 info!("Starting IOTA Names Indexer");
 
                 let prometheus = PrometheusServer::new();
                 let registry = prometheus.registry();
+                let metrics = Arc::new(IotaNamesMetrics::new(&registry));
+
+                // Try to restore metrics from Prometheus
+                if let Err(e) = metrics.restore_from_prometheus(&prometheus_url).await {
+                    info!("Could not restore metrics from Prometheus ({e}), starting fresh");
```
thibault-martinez: Maybe a warning instead of info?

thibault-martinez: Also, do we really start fresh? What if it fails in the middle of restoring them? Wouldn't some of the metrics be filled and some not?

Thoralf-M: Yeah, changed the message https://github.com/iotaledger/iota-names/pull/288/commits/490dd9596dac8e88969862aa7293b0cae100c9d9

indexer/docker/docker-compose.yml
Thoralf-M: It's another service in the docker-compose file, localhost wouldn't work as it would look for something inside of the iota-names-indexer with this port

indexer/src/metrics.rs
thibault-martinez, line 1: `auction_bid_count_distribution` is missing

thibault-martinez (changes requested):

indexer/src/metrics.rs
Thoralf-M: This doesn't even exist?

thibault-martinez: https://github.com/iotaledger/iota-names/blob/27d69e6c5197939b6179ce239e9469c7dc4d8b5e/indexer/src/metrics.rs#L33 ?

Thoralf-M: Ah well develop wasn't merged

Thoralf-M: Done here https://github.com/iotaledger/iota-names/pull/288/commits/ebded09b1a2785df87a23b029d8d867d12d1dad2

thibault-martinez, line 1: There is nothing restorable from histograms, right?

Thoralf-M: Seems like one can do it only in a limited/approximated way, I haven't tried it so not sure how good the result would be, but it also seems quite complex even for that, so I thought it's not worth it

indexer/src/metrics.rs
```diff
@@ -157,6 +157,213 @@ impl IotaNamesMetrics {
             .unwrap(),
         }
     }
+
+    /// Restore metrics from Prometheus by querying the latest values
+    pub async fn restore_from_prometheus(&self, prometheus_url: &str) -> anyhow::Result<()> {
```
thibault-martinez, line 162: An idea for later, no need to do it now: I would have merged `IotaNamesMetrics::new` and `IotaNamesMetrics:: restore_from_prometheus` and merged the register and restore macros so it's impossible to miss restoring a metric in the future, because declaring it would also restore it

thibault-martinez (changes requested):
Indexer doesn't compile

indexer/src/metrics.rs
```diff
@@ -157,6 +157,213 @@ impl IotaNamesMetrics {
...
+
+        if let Some(results) = response["data"]["result"].as_array() {
+            for result in results {
+                if let (Some(labels), Some(value_str)) =
+                    (result["metric"].as_object(), result["value"][1].as_str())
+                {
+                    let value = value_str.parse::<f64>()? as i64;
+
+                    if let Some(label_value) = labels.get(label_key).and_then(|v| v.as_str()) {
+                        match metric_name {
```
thibault-martinez: I think this is a bit weird. Why don't you do like `query_metric_value`? Return the value and set it in the macro?

thibault-martinez (approved):

indexer/src/metrics.rs
thibault-martinez: Way better now :)

PR #291, opened 2025-06-16
feat(metrics): add auction house balance metric

### LINKED ISSUE #7138: IOTA-Names metrics: balance of AuctionHouse

Alex6323:
fixes https://github.com/iotaledger/iota/issues/7138

image


DaughterOfMars:
Shouldn't we just redefine the config env vars in this repo? I don't think we should add IDs to the core one but it makes sense for this crate to just manually define them all since it doesn't have a client.

packages/auction/sources/auction.move
```diff
@@ -50,10 +55,14 @@ const ENoProfits: vector<u8> = b"There are no profits to withdraw.";
 /// Authorization witness to call protected functions of `iota_names`.
 public struct AuctionAuth has drop {}
 
+public struct BalanceKey<phantom T> has copy, drop, store {}
+
 /// The AuctionHouse application.
+/// 
+/// Dynamic fields:
+/// - `balance: BalanceKey<IOTA> -> Balance<IOTA>`
```
Thoralf-M: What's the motivation to put this in a dynamic field if we only use the IOTA coin anyways? In the main package it's different, since we might want to accept other coins in the future, but that's not the case for the auction

Alex6323: Out-of-interest: why would we not accept that for the auction as well at some point?

thibault-martinez: Auctions are temporary, they will probably only last for a month or two, we'll never have/allow another coin by then

Alex6323: Not sure it's the best approach, but the motivation for making the balance a dynamic field, was that I could easily access it without overfetching.

thibault-martinez: You had the auction house through the ID but you moved the balance away so isn't that actually more fetches/computations?

thibault-martinez: In any case, I wouldn't complicate the SCs for the sake of the indexer tbh

Alex6323: yeah, but wouldn't I have to fetch the full auction house bcs blob then to get access to the balance?

thibault-martinez: I think it's completely negligible

Alex6323: if there wasn't a table that can grow in size I would agree, but to always deserialize the whole thing when we just want the balance seems inefficient to me. Are there compelling reasons for not using a dynamic field?

Alex6323: @Thoralf-M @DaughterOfMars wdyt?

thibault-martinez: But you don't have to deserialise the whole table

Alex6323:
> Shouldn't we just redefine the config env vars in this repo? I don't think we should add IDs to the core one but it makes sense for this crate to just manually define them all since it doesn't have a client.

I guess you're right. It felt weird to me as well, but I wasn't sure. Thanks!

packages/auction/sources/auction.move
Thoralf-M: Table is built on top of dynamic fields, so you don't need to deserialize the whole thing to deserialize the AuctionHouse
> Are there compelling reasons for not using a dynamic field?

It's simpler in the smart contract and also easier fetch the balance, since you don't have to fetch a dynamic field

Alex6323: thanks guys

Thoralf-M: Also need to undo the comment

scripts/package-info/devnet.json
```diff
@@ -2,6 +2,7 @@
     "adminAddress": "0xb98bc58cd2cae4c621b47ed0a3bbed2e1dcb78da60d578625384dad3a231c6fb",
     "adminCap": "0x84096eeb81f666bc8e7b14803b387617ad92aa796b5f84ddc752a513b273fc34",
     "auctionPackageId": "0x5e7a300e640f645a4030aeb507c7be16909e6fa9711e7ca2d4397bbd967d5c50",
+    "auctionHousseId": "0x12345",
```
Thoralf-M:
```suggestion
    "auctionHouseId": "0x31deb8cbd320867089d52c37fed2d443520aac0fc5a957de1f64f9135b83f42b",
```

thibault-martinez:
Please add a screenshot when you are done 🙏🏻

indexer/src/worker.rs
```diff
@@ -69,6 +71,7 @@ pub(crate) async fn run_iota_names_reader(
 
 pub(crate) struct IotaNamesWorker {
     config: IotaNamesConfig,
+    extended_config: IotaNamesExtendedConfig,
```
thibault-martinez, line 80: Can we maybe put the normal one in the extended?

scripts/package-info/devnet.json
```diff
@@ -2,6 +2,7 @@
     "adminAddress": "0xb98bc58cd2cae4c621b47ed0a3bbed2e1dcb78da60d578625384dad3a231c6fb",
     "adminCap": "0x84096eeb81f666bc8e7b14803b387617ad92aa796b5f84ddc752a513b273fc34",
     "auctionPackageId": "0x5e7a300e640f645a4030aeb507c7be16909e6fa9711e7ca2d4397bbd967d5c50",
+    "auctionHouseId": "0x31deb8cbd320867089d52c37fed2d443520aac0fc5a957de1f64f9135b83f42b",
```
thibault-martinez: There seems to be a little bit of inconsistency, here we have `auctionHouseId` but `iotaNames`. In another file we have `iotaNames` and `auctionHouse`

Alex6323: Yeah, also noticed that. Do you prefer it with the ID suffix or without. I would change to whatever you prefer. I honestly don't care that much.

Thoralf-M (approved):

scripts/package-info/devnet.json
Thoralf-M: https://github.com/iotaledger/iota-names/pull/291/commits/6847bf93d3b366657c74ceea605b621ee93498ca

indexer/src/worker.rs
Thoralf-M: https://github.com/iotaledger/iota-names/pull/291/commits/6847bf93d3b366657c74ceea605b621ee93498ca

DaughterOfMars (approved):

scripts/utils/envs.ts
```diff
@@ -24,10 +24,11 @@ const json = JSON.parse(fs.readFileSync(filePath, 'utf-8'));
 
 const envVars = {
     IOTA_NAMES_PACKAGE_ADDRESS: json.packageId,
-    IOTA_NAMES_OBJECT_ID: json.iotaNames,
+    IOTA_NAMES_OBJECT_ID: json.iotaNamesObjectId,
     IOTA_NAMES_PAYMENTS_PACKAGE_ADDRESS: json.paymentsPackageId,
     IOTA_NAMES_REGISTRY_ID: json.registryTableId,
     IOTA_NAMES_REVERSE_REGISTRY_ID: json.reverseRegistryTableId,
+    IOTA_NAMES_AUCTION_HOUSE_ID: json.auctionHouseObjectId,
```
thibault-martinez: Should be `IOTA_NAMES_AUCTION_HOUSE_OBJECT_ID` for consistency then?

Thoralf-M: https://github.com/iotaledger/iota-names/pull/291/commits/7780eefad400d319e4c6fa44cd07bf4fbfbe06d0

thibault-martinez (approved):

evavirseda (approved):

PR #292, opened 2025-06-16
feat: add average-of-bids-per-auction metric

### LINKED ISSUE #7137: IOTA-Names metrics: average number of bid per auction

Alex6323:
fixes https://github.com/iotaledger/iota/issues/7137

image


indexer/src/worker.rs
```diff
@@ -161,14 +168,35 @@ impl IotaNamesWorker {
...
             IotaNamesEvent::AuctionExtended(_event) => (),
             IotaNamesEvent::AuctionFinalized(event) => {
                 self.metrics.total_auction_finalized.inc();
                 self.metrics.auction_final_prices.observe(event.winning_bid);
+                let bid_count = self
+                    .bids_per_domain
+                    .lock()
+                    .expect("error taking lock")
+                    .remove(&event.domain.to_string())
+                    .expect("missing auction");
```
Thoralf-M: We shouldn't panic here and in AuctionBid if an auction is missing, if we restart the indexer this might very well happen and shouldn't stop the indexer
I would just ignore this case then

Alex6323: yeah

Alex6323: https://github.com/iotaledger/iota-names/pull/292/commits/e494f1056f426ef9a2619cae4053f70becfc6e0b

Thoralf-M (approved):

indexer/src/worker.rs
```diff
@@ -72,6 +77,7 @@ pub(crate) struct IotaNamesWorker {
     metrics: Arc<IotaNamesMetrics>,
     token: CancellationToken,
     balance_object_id: ObjectID,
+    bids_per_domain: Arc<Mutex<HashMap<String, u64>>>,
```
DaughterOfMars: 🤔 This should probably use the database.

Alex6323: not sure about that, because if we keep it in memory, we can be sure the result is correct. When we use the database, however, i.e. persist it, we might not know whether we've seen all bids in case the indexer was down in between?

thibault-martinez: How can the indexer miss bids since it always starts where it stopped?

Alex6323: switching to use the database ... we still remove finished auctions from the table, right?

indexer/diesel.toml
```diff
@@ -1,2 +1,3 @@
-[print_schema]
-file = "src/db/models.rs"
+# TODO do we still need this, because it's overwriting our manually written src/db/models.rs?
```
thibault-martinez: I don't think it's actually touching this file? It just reads it to generate migrations etc

indexer/migrations/2025-06-13-134255_create_auctions/down.sql
```diff
@@ -2,3 +2,4 @@
 DROP TABLE IF EXISTS `domains`;
 DROP TABLE IF EXISTS `bidders`;
 DROP TABLE IF EXISTS `bidder_domain`;
+DROP TABLE IF EXISTS `domain_bids`;
```
thibault-martinez: I would personally create a new migration, not 100% necessary since we're not running yet but I would start getting into the habit of doing it

indexer/src/worker.rs
```diff
@@ -187,6 +193,9 @@ impl IotaNamesWorker {
                         &event.domain.to_string(),
                     )
                 })?;
+                conn.transaction::<_, anyhow::Error, _>(|conn| {
```
thibault-martinez: You should be able to reuse the previous transaction

indexer/diesel.toml
Thoralf-M: Yes it's just reading from there

Alex6323: no, it's not

```
// @generated automatically by Diesel CLI.

diesel::table! {
    bidder_domain (bidder_id, domain_id) {
        bidder_id -> Integer,
        domain_id -> Integer,
    }
}

diesel::table! {
    bidders (id) {
        id -> Nullable<Integer>,
        address -> Text,
    }
}

diesel::table! {
    domain_bids (domain_id) {
        domain_id -> Integer,
        bids -> Integer,
    }
}

diesel::table! {
    domains (id) {
        id -> Nullable<Integer>,
        name -> Text,
    }
}

diesel::joinable!(bidder_domain -> bidders (bidder_id));
diesel::joinable!(bidder_domain -> domains (domain_id));

diesel::allow_tables_to_appear_in_same_query!(
    bidder_domain,
    bidders,
    domain_bids,
    domains,
);

```

Alex6323: `diesel database reset` for example changes it

indexer/migrations/2025-06-13-134255_create_auctions/down.sql
Alex6323: okay

indexer/diesel.toml
Thoralf-M: But why reset it?

Alex6323: was just the first example that came to mind: `diesel migration run` also re-generates it.

Thoralf-M: We just used `diesel migration generate --database-url AUCTIONS_DB --diff-schema create_auctions` and it didn't touch this file, why do we need anything else? (apart from another name than `create_auctions`)

thibault-martinez: We'll probably never use the diesel CLI

indexer/src/worker.rs
Alex6323: https://github.com/iotaledger/iota-names/pull/292/commits/1869ad8094416a46f07495b6a38df278efa10ee1

indexer/src/db/queries.rs
```diff
@@ -84,3 +84,37 @@ pub fn get_domains_for_bidder_address(
...
 }
+
+pub fn add_domain_bids_entry(conn: &mut SqliteConnection, domain_name: &str) -> Result<()> {
+    let domain = get_or_create_domain(conn, domain_name)?;
+    insert_into(domain_bids::table)
+        .values((
+            domain_bids::domain_id.eq(domain.id),
+            domain_bids::bids.eq(1),
+        ))
+        .on_conflict_do_nothing()
```
DaughterOfMars: Shouldn't we be overwriting?

DaughterOfMars: And if we do then we can remove the update method. Blind inserts are probably faster anyway.

Alex6323: https://github.com/iotaledger/iota-names/pull/292/commits/147b22b5f8e2b448c2dbeec7fd19eeca230daa69

indexer/src/worker.rs
```diff
@@ -219,7 +232,6 @@ impl IotaNamesWorker {
                         .inc();
                 }
             }
-            // `subdomains`
```
Thoralf-M: Why remove? There are also comments for other things above

Alex6323: sorry, that was accidentally removed during merge

DaughterOfMars (approved):

Thoralf-M (approved):

indexer/src/worker.rs
```diff
@@ -205,6 +218,7 @@ impl IotaNamesWorker {
                         .inc();
                 }
             }
+            // `subdomains`
```
thibault-martinez: Need to put it back where it was

indexer/migrations/2025-06-18-131642_create_domain_bids/up.sql
```diff
@@ -0,0 +1,6 @@
+-- Your SQL goes here
+CREATE TABLE `domain_bids`(
+	`domain_id` INTEGER NOT NULL,
+	`bids` INTEGER NOT NULL,
+	PRIMARY KEY(`domain_id`)
```
thibault-martinez: Do we not need a
```
FOREIGN KEY (`domain_id`) REFERENCES `domains`(`id`)
```
like we do for `bidder_domain`?

indexer/src/db/models.rs
```diff
@@ -48,7 +56,14 @@ diesel::table! {
...
+        domain_id -> Int4,
+        bids -> Int4,
+    }
+}
+
 diesel::joinable!(bidder_domain -> domains (domain_id));
 diesel::joinable!(bidder_domain -> bidders (bidder_id));
 
-diesel::allow_tables_to_appear_in_same_query!(bidders, domains, bidder_domain);
+diesel::allow_tables_to_appear_in_same_query!(bidders, domains, bidder_domain, domain_bids);
```
thibault-martinez: Do we really need it here?

indexer/src/worker.rs
```diff
@@ -128,7 +131,8 @@ impl IotaNamesWorker {
                         conn,
                         &event.bidder.to_string(),
                         &event.domain.to_string(),
-                    )
+                    )?;
+                    upsert_domain_bids_entry(conn, &event.domain.to_string())
```
thibault-martinez: Maybe convert it to string only once for the 2 queries

indexer/src/worker.rs
```diff
@@ -138,7 +142,8 @@ impl IotaNamesWorker {
...
```
thibault-martinez: same

indexer/src/worker.rs
Alex6323: arrrrgh

indexer/migrations/2025-06-18-131642_create_domain_bids/up.sql
Alex6323: I think it's not required, otherwise I would have run into issues. We could add it though, if we wanted to make sure that the `domains` table has been updated before. There's not a real benefit of it IMO though, unless it would be interesting to query for the domains that are currently being observed for bids.

indexer/src/db/models.rs
Alex6323: actually with the former comment it shouldn't be added here

indexer/migrations/2025-06-18-131642_create_domain_bids/up.sql
Alex6323: Gave it another thought and I think it's cleaner to actually add it even if we don't really need the connection between the tables.

indexer/src/db/models.rs
Alex6323: https://github.com/iotaledger/iota-names/pull/292/commits/19cba1d01c5492a250edff4f57726f438f3cc034

indexer/migrations/2025-06-18-131642_create_domain_bids/up.sql
Alex6323: https://github.com/iotaledger/iota-names/pull/292/commits/19cba1d01c5492a250edff4f57726f438f3cc034

indexer/src/worker.rs
Alex6323: https://github.com/iotaledger/iota-names/pull/292/commits/3aefad84cccfec132b1ad608acf428b281a8c64f

indexer/src/worker.rs
Alex6323: https://github.com/iotaledger/iota-names/pull/292/commits/3aefad84cccfec132b1ad608acf428b281a8c64f

thibault-martinez (approved):

PR #304, opened 2025-06-20
feat(iota-names): add calculate_base_price_with_domain()

### LINKED ISSUE #189: Change `pricing_config::calculate_base_price` to accept a `Domain` (or a string)
## Description

The `calculate_base_price` function takes a length, but this is very inconvenient to call and may be more useful if you could pass a `Domain` or possibly a string.

```
public fun calculate_base_price(config: &PricingConfig, length: u64): u64
```

Thoralf-M:
Add calculate_base_price_with_domain() so one can provide a domain and doesn't has to compute the SLD length

Closes https://github.com/iotaledger/iota-names/issues/189

### DIFF
--- a/packages/iota-names/sources/sales/pricing_config.move
+++ b/packages/iota-names/sources/sales/pricing_config.move
@@ -43,9 +43,9 @@ public fun calculate_base_price(config: &PricingConfig, length: u64): u64 {
     *config.pricing.get(&range)
 }
 
-/// Calculates the base price for a given domain based on its SLD length.
+/// Calculates the base price of a given domain based on its SLD length.
 /// For example, "my-name.iota" uses length 7 (the length of "my-name").
-public fun calculate_base_price_with_domain(config: &PricingConfig, domain: Domain): u64 {
+public fun calculate_base_price_of_domain(config: &PricingConfig, domain: Domain): u64 {
     calculate_base_price(config, domain.sld().length())
 }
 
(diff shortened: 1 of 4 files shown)


packages/iota-names/sources/sales/pricing_config.move
```diff
@@ -43,6 +43,12 @@ public fun calculate_base_price(config: &PricingConfig, length: u64): u64 {
     *config.pricing.get(&range)
 }
 
+/// Calculates the base price for a given domain based on its SLD length.
+/// For example, "my-name.iota" uses length 7 (the length of "my-name").
+public fun calculate_base_price_with_domain(config: &PricingConfig, domain: Domain): u64 {
```
thibault-martinez: Not a strong opinion but I feel like `with_domain` is a bit weird

thibault-martinez (approved):

packages/iota-names/sources/sales/pricing_config.move
Thoralf-M: Any other idea? Since without `_with_domain` is already used

thibault-martinez: `of_domain` or `for_domain` sound better to me but maybe a native speaker would know better @DaughterOfMars

DaughterOfMars: I agree, either would be fine

Thoralf-M: https://github.com/iotaledger/iota-names/pull/304/commits/eeec3f6865c7cf75be8fc2a86dc16d91cb8a3342

DaughterOfMars (approved):

PR #311, opened 2025-06-20
feat(indexer): check event package ID

### LINKED ISSUE #244: Revert allowing events from all packages
So is this function now failing for non-IOTA-Names events?

_Originally posted by @thibault-martinez in https://github.com/iotaledger/iota-names/pull/236#discussion_r2123373212_
            
We should only consider events coming from IOTA-Names packages

thibault-martinez:
Fixes https://github.com/iotaledger/iota-names/issues/244


indexer/src/events.rs
```diff
@@ -32,43 +34,61 @@ pub(crate) enum IotaNamesEvent {
...
-            _ => anyhow::bail!("Invalid event type: {}", event.type_.name),
-        }))
+        let package_address = IotaAddress::from(event.package_id);
+
+        if package_address == config.auction_package_address
+            || package_address == config.coupons_package_address
+            || package_address == config.iota_names_config.package_address
+            || package_address == config.iota_names_config.payments_package_address
+            || package_address == config.subdomains_package_address
+        {
```
Thoralf-M: Would be a bit nicer to just early return with None here if an Event has another package id, instead of having everything in this closure

thibault-martinez: https://github.com/iotaledger/iota-names/pull/311/commits/75e7221954a448bb22694b2a1f54cf0552c6cd0a

indexer/src/events.rs
```diff
@@ -32,9 +34,19 @@ pub(crate) enum IotaNamesEvent {
...
+        let package_address = IotaAddress::from(event.package_id);
+
+        if package_address != config.auction_package_address
+            && package_address != config.coupons_package_address
+            && package_address != config.iota_names_config.package_address
+            && package_address != config.iota_names_config.payments_package_address
+            && package_address != config.subdomains_package_address
+        {
+            return Ok(None);
+        }
```
Thoralf-M: How about this?
```suggestion
        let allowed_packages = [
            config.auction_package_address,
            config.coupons_package_address,
            config.iota_names_config.package_address,
            config.iota_names_config.payments_package_address,
            config.subdomains_package_address,
        ];

        if !allowed_packages.contains(&package_address) {
            return Ok(None);
        }
```

Thoralf-M: Or even add a method to the extended config

thibault-martinez: I can add a method to the config yeah, but I would probably still do ifs to avoid building an array for every single event

thibault-martinez: https://github.com/iotaledger/iota-names/pull/311/commits/28d75632ff8f62630b9c808c9dac457d592c42e1

Thoralf-M (approved):

Alex6323 (approved):

PR #325, opened 2025-06-24
feat: add bidding option when searching for a name

### LINKED ISSUE #305: Add bidding option when searching for a name
When the user searches for a name, if the name is available, we should allow them to also bid, so that we can test auctions.

This search and option suggested to the user to bid should include both available names that dont have any auction yet so the user will be the first one, and also names that already are in an auction that is not ended. Note that the price displayed should be the minimum bid the user has to place in order to be able to succesfully bid, which for active auctions is highest big +1 IOTA (I think)

Make sure to not include auctions that are already ended, we should offer the option to bid only when possible, that means available names with no auction, and names with an auction that has not yet ended.

I would also add next to the price, if possible, when the auction ends (if there is an live auction)

⚠ This task is not about the bidding action, is only about fetching and displaying prices and a button

Image

evavirseda:
for a current auction search for: `test1` or `test`

fixes https://github.com/iotaledger/iota-names/issues/305


dapp/src/components/AvailabilityCheck.tsx
```diff
@@ -38,25 +39,42 @@ function getValidationError(
...
     const [searchValue, setSearchValue] = useState<string>('');
     const [name, setName] = useState<string>('');
     const [isPurchaseDialogOpen, setPurchaseDialogOpen] = useState(false);
 
-    const { data, error } = useNameRecord(name);
+    const { data: nameRecordData, error } = useNameRecord(name);
     const { data: priceList } = usePriceList();
 
+    const isAuctionInProgress =
+        nameRecordData?.type === 'unavailable' && !!nameRecordData?.nameRecord;
```
VmMad: I think this logic is incorrect. If it's unavailable and the nameRecord exists, that means it's already bought, right??

VmMad: image

dapp/src/components/AvailabilityCheck.tsx
VmMad, line 1: There's also a little flash, when you search for the `test` name, and then you search for `test1` you can see briefly the 1 IOTA price, which then changes to 2. If possible it would be good to show a '--' or a loader if the price is not finished laoding

dapp/src/components/AvailabilityCheck.tsx
```diff
@@ -69,13 +84,34 @@ export function AvailabilityCheck() {
         setName('');
     }
 
-    const errorMessage = error?.message ?? validationError ?? '';
-    const canBuy = data?.type === 'available';
-    const enableSearch = Boolean(searchValue) && !errorMessage;
+    const statusLabel = useMemo(() => {
```
marc2332: No need for the use memo here, this is not expensive to execute

marc2332 (changes requested):

marc2332 (approved):

dapp/src/components/AvailabilityCheck.tsx
```diff
@@ -38,25 +39,39 @@ function getValidationError(
     return null;
 }
 
+const ONE_IOTA_NANOS = 1_000_000_000;
```
msarcev: There are constants in ts-sdk for this, lets use those

msarcev: `NANOS_PER_IOTA`

dapp/src/components/AvailabilityCheck.tsx
```diff
@@ -69,13 +84,34 @@ export function AvailabilityCheck() {
...
+        if (isAvailable) {
+            return <span className="text-green-700 dark:text-green-200">Available</span>;
+        } else if (nameRecordData?.type === 'not-priced') {
+            return <span className="text-red-700 dark:text-red-200">Not priced</span>;
+        } else if (nameRecordData?.type === 'unavailable') {
+            if (isAuctionMetadataLoading) {
+                return <Skeleton widthClass="w-32" heightClass="h-6" />;
+            }
+
+            const isInProgress =
```
msarcev: Can we reuse the bool computed at L57 for this ?

msarcev (changes requested):

msarcev (approved):

PR #326, opened 2025-06-24
chore(indexer): Slight improvements to startup code

DaughterOfMars:
## Description

A little cleanup for the startup logic in the indexer.


indexer/src/main.rs
```diff
@@ -124,15 +115,18 @@ impl Command {
                         }
                     },
                     res = tasks.join_next() => {
+                        cancel_token.cancel();
+                        tracing::debug!("tasks have begun shutting down");
                         if let Some(Ok(Err(err))) = res {
                             tracing::error!("a worker failed with error: {err}");
                             exit_code = Err(err);
                         }
+                        while let Some(Ok(Err(err))) = tasks.join_next().await {
```
thibault-martinez: What if one of them doesn't fail? This would stop the loop and not show errors for other worker that may fail? Is this intended?

DaughterOfMars: Ah, good catch

indexer/src/main.rs
```diff
@@ -124,15 +115,20 @@ impl Command {
...
+                        tracing::debug!("tasks have begun shutting down");
                         if let Some(Ok(Err(err))) = res {
                             tracing::error!("a worker failed with error: {err}");
                             exit_code = Err(err);
                         }
+                        while let Some(res) = tasks.join_next().await {
+                            if let Ok(Err(err)) = res {
+                                tracing::error!("a worker failed with error: {err}");
+                            }
+                        }
```
thibault-martinez, line 128: Feels like there is a way to combine the if and when by resetting `res` but no big deal

thibault-martinez (approved):

Thoralf-M (approved):

indexer/src/main.rs
DaughterOfMars: 🤷‍♀️  I think it's fine

thibault-martinez: All good, I approved anyway

PR #334, opened 2025-06-25
feat(indexer): use remote_store to sync

### LINKED ISSUE #333: Upgrade indexer to new CheckpointReader
https://github.com/iotaledger/iota/pull/7424

Thoralf-M:
Make use of the indexer reader v2 with remote store, like `https://checkpoints.devnet.iota.cafe`
For localnets I kept the old version with REST API

To skip syncing the big genesis and all the checkpoints before we even published the package, the `progress_store` should be set to something like
```JSON
{
    "iota_names_reader": 67263904
}
```
where the checkpoint is just before the package was published

I changed the COUPONS_PACKAGE_ADDRESS because it would otherwise panic if no env values are provided
Also exposed the 3030 port for auctions

Fixes https://github.com/iotaledger/iota-names/issues/333

Tested with `RUST_LOG=debug,hyper_util=off cargo run start --node-url https://checkpoints.devnet.iota.cafe`
and `docker compose -f indexer/docker/docker-compose.yml up`

### DIFF
--- a/indexer/docker/docker-compose.yml
+++ b/indexer/docker/docker-compose.yml
@@ -14,7 +14,7 @@ services:
     image: iota-names-indexer:dev
     ports:
       - "9189:9189/tcp" # Metrics
-      - 3030:3030 # Auction
+      - 3030:3030 # Auction REST API
     tty: true
     deploy:
       restart_policy:
--- a/indexer/docker/docker-compose.yml
+++ b/indexer/docker/docker-compose.yml
@@ -14,6 +14,7 @@ services:
     image: iota-names-indexer:dev
     ports:
       - "9189:9189/tcp" # Metrics
+      - 3030:3030 # Auction
     tty: true
     deploy:
       restart_policy:
--- a/indexer/src/worker.rs
+++ b/indexer/src/worker.rs
@@ -64,7 +64,7 @@ pub(crate) async fn run_iota_names_reader(
 
     info!("Connecting to node at {node_url}");
     // Localnet does not support remote store, so we use the REST API
-    if node_url == "http://localhost:9000" || node_url == "http://host.docker.internal:9000" {
+    if node_url.contains("localhost") || node_url.contains("host.docker.internal") {
         executor
             .run(
                 // path to a local directory where checkpoints are stored.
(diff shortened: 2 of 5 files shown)


indexer/docker/docker-compose.yml
```diff
@@ -14,6 +14,7 @@ services:
     image: iota-names-indexer:dev
     ports:
       - "9189:9189/tcp" # Metrics
+      - 3030:3030 # Auction
```
thibault-martinez:
```suggestion
      - 3030:3030 # REST API
```

indexer/src/worker.rs
```diff
@@ -62,16 +63,28 @@ pub(crate) async fn run_iota_names_reader(
...
-            // path to a local directory where checkpoints are stored.
-            PathBuf::from("./chk"),
-            Some(format!("{node_url}/api/v1")),
-            // optional remote store access options.
-            vec![],
-            ReaderOptions::default(),
-        )
-        .await?;
+    // Localnet does not support remote store, so we use the REST API
+    if node_url == "http://localhost:9000" || node_url == "http://host.docker.internal:9000" {
```
thibault-martinez: Is the port not configurable?

indexer/docker/docker-compose.yml
Thoralf-M: Well technically the metrics are also provided by a REST API

indexer/src/worker.rs
Thoralf-M: It is, but who changes it? I can use contains() with "localhost" and "host.docker.internal" if you prefer that

indexer/docker/docker-compose.yml
thibault-martinez: Sure sure, but the readme differentiate between Metrics and REST API. I would just mention auctions here

indexer/src/worker.rs
thibault-martinez: I think it would be slightly better yeah

Thoralf-M: https://github.com/iotaledger/iota-names/pull/334/commits/b066a746fd564e09f45bc72d9f5f4b729f824f19

indexer/docker/docker-compose.yml
Thoralf-M: Also updated the README https://github.com/iotaledger/iota-names/pull/334/commits/b066a746fd564e09f45bc72d9f5f4b729f824f19

DaughterOfMars (approved):

indexer/docker/docker-compose.yml
thibault-martinez: I would personally have kept it generic in case we add more things than auctions but that's fine for now

thibault-martinez (approved):

PR #335, opened 2025-06-25
feat(iota-names): add admin_remove_records

### LINKED ISSUE #330: Add function to remove record from Registry

Thoralf-M:
Add a function for the admin to force remove records by their domain name

Can be used like this
```
ADMIN_CAP=0x03c2f9b2a3dbbfa50827a65b2003883e0d8d5c170d76166226faa99ecb58483b
IOTA_NAMES_PACKAGE_ID=0x921f4ce634cc7222b42aab7c2b16a2eb80128d3ac53fdff077c115f8f3d80fc7
IOTA_NAMES_OBJECT_ID=0x2e6a383f27cc9dfa05ef088b5a97b3d3e60bd3ee80d8d94f59eb6ece55c6e01c
DOMAIN_RECORDS_TO_REMOVE='["test.iota", "existing1.iota"]'
iota client ptb \
--make-move-vec "<std::string::String>" $DOMAIN_RECORDS_TO_REMOVE \
--assign domains_to_remove \
--move-call $IOTA_NAMES_PACKAGE_ID::admin::admin_remove_records @$ADMIN_CAP @$IOTA_NAMES_OBJECT_ID domains_to_remove \
--dry-run
```

Closes https://github.com/iotaledger/iota-names/issues/330

### DIFF
--- a/packages/iota-names/sources/admin.move
+++ b/packages/iota-names/sources/admin.move
@@ -66,6 +66,7 @@ public fun admin_remove_records(
     mut domains: vector<String>,
 ) {
     let registry = iota_names::auth_registry_mut<AdminAuth, Registry>(AdminAuth {}, iota_names);
+    assert!(!domains.is_empty(), ENoDomainsProvided);
     while (!domains.is_empty()) {
         let domain = domain::new(domains.pop_back());
         registry.admin_force_remove_record(domain);
--- a/packages/iota-names/sources/admin.move
+++ b/packages/iota-names/sources/admin.move
@@ -52,3 +52,18 @@ entry fun reserve_domains(
         iota::transfer::public_transfer(nft, sender);
     };
 }
+
+/// Admin function to forcefully remove registry entries by their domain.
+/// This bypasses all expiration and authorization checks and immediately 
+/// removes the records and invalidates any reverse lookup entries.
+public fun admin_remove_records(
+    _: &AdminCap,
+    iota_names: &mut IotaNames,
+    mut domains: vector<String>,
+) {
+    let registry = iota_names::auth_registry_mut<AdminAuth, Registry>(AdminAuth {}, iota_names);
+    while (!domains.is_empty()) {
+        let domain = domain::new(domains.pop_back());
+        registry.admin_force_remove_record(domain);
+    };
+}
--- a/packages/iota-names/sources/admin.move
+++ b/packages/iota-names/sources/admin.move
@@ -69,6 +69,6 @@ public fun admin_remove_records(
     assert!(!domains.is_empty(), ENoDomainsProvided);
     while (!domains.is_empty()) {
         let domain = domain::new(domains.pop_back());
-        registry.admin_force_remove_record(domain);
+        registry.force_remove_record(domain);
     };
 }
--- a/packages/iota-names/sources/admin.move
+++ b/packages/iota-names/sources/admin.move
@@ -15,6 +15,9 @@ use iota_names::iota_names_registration::IotaNamesRegistration;
 use iota_names::registry::Registry;
 use std::string::String;
 
+#[error]
+const ENoDomainsProvided: vector<u8> = b"No domains provided";
+
 /// Authorization witness to call protected functions of `iota_names`.
 public struct AdminAuth has drop {}
 
@@ -45,6 +48,7 @@ entry fun reserve_domains(
     let sender = sender(ctx);
     let config = *iota_names.get_config<CoreConfig>();
     let registry = iota_names::auth_registry_mut<AdminAuth, Registry>(AdminAuth {}, iota_names);
+    assert!(!domains.is_empty(), ENoDomainsProvided);
     while (!domains.is_empty()) {
         let domain = domain::new(domains.pop_back());
         config.assert_is_valid_for_sale(&domain);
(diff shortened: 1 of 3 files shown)


packages/iota-names/sources/admin.move
```diff
@@ -52,3 +52,18 @@ entry fun reserve_domains(
...
+/// Admin function to forcefully remove registry entries by their domain.
+/// This bypasses all expiration and authorization checks and immediately 
+/// removes the records and invalidates any reverse lookup entries.
+public fun admin_remove_records(
+    _: &AdminCap,
+    iota_names: &mut IotaNames,
+    mut domains: vector<String>,
+) {
+    let registry = iota_names::auth_registry_mut<AdminAuth, Registry>(AdminAuth {}, iota_names);
+    while (!domains.is_empty()) {
```
DaughterOfMars, line 70: Should we throw an error if the domains list is empty?

packages/iota-names/sources/registry.move
```diff
@@ -239,6 +239,26 @@ public fun remove_leaf_record(self: &mut Registry, domain: Domain) {
     self.handle_invalidate_reverse_record(&domain, old_target_address, none());
 }
 
+/// Admin function to forcefully remove any record by domain.
+/// This bypasses all expiration and authorization checks.
+/// Should only be called by admin functions in the main module.
+public fun admin_force_remove_record(self: &mut Registry, domain: Domain) {
```
DaughterOfMars: Maybe for now make this package level?

packages/iota-names/sources/registry.move
```diff
@@ -239,6 +239,26 @@ public fun remove_leaf_record(self: &mut Registry, domain: Domain) {
...
+    // Check if the record exists before trying to remove it
+    if (!self.registry.contains(domain)) {
+        return
```
DaughterOfMars, line 248: Maybe an error?

Thoralf-M: I thought it might be better that `admin_remove_records()` doesn't fail just because one record is missing, from the time one prepared a list until it's executed it might be deleted in between already
But I guess the chance for that is pretty low and one can do a dry run before to see if would work

packages/iota-names/sources/registry.move
Thoralf-M: What do you mean with package level?

packages/iota-names/sources/admin.move
Thoralf-M: Changed in https://github.com/iotaledger/iota-names/pull/335/commits/e35fd734b2ca950c4d9a25cf3c3f70c3e58a901e

thibault-martinez: You've added it to another function?

packages/iota-names/sources/registry.move
DaughterOfMars: I mean not full public

packages/iota-names/sources/registry.move
DaughterOfMars: Ah yeah it's unfortunate we can't just handle the result :c

packages/iota-names/sources/admin.move
Thoralf-M: Oh right, fixed

packages/iota-names/sources/registry.move
Thoralf-M: Needs to be pub, otherwise we can't call it from the other module
But all the registry methods are protected because you can only call them if you have an authorized package to get the registry in the first place

DaughterOfMars: Can it not be `public(package)`?

Thoralf-M: Right that works, changed here https://github.com/iotaledger/iota-names/pull/335/commits/4ca86c09c3f01c9de68ff5d0fe70fd019e95a88f

DaughterOfMars (approved):

packages/iota-names/sources/registry.move
```diff
@@ -239,6 +239,26 @@ public fun remove_leaf_record(self: &mut Registry, domain: Domain) {
     self.handle_invalidate_reverse_record(&domain, old_target_address, none());
 }
 
+/// Admin function to forcefully remove any record by domain.
+/// This bypasses all expiration and authorization checks.
+/// Should only be called by admin functions in the main module.
+public(package) fun admin_force_remove_record(self: &mut Registry, domain: Domain) {
```
thibault-martinez: So the fact it's `public(package)` makes it admin? Otherwise if it takes a `&mut Registry` it's auth, not admin?

Thoralf-M: Nothing here makes it admin, will rename it

Thoralf-M: https://github.com/iotaledger/iota-names/pull/335/commits/31462a7598ce971adc32767131415655037d29ad

thibault-martinez: Ok but it definitely can't be called by authorised apps because of `public(package)`?

Thoralf-M: Yes (if you don't count the iota-names modules as an authorized app), unless one adds another wrapper function first

thibault-martinez: "one" being us?

Thoralf-M: Yes, owner of the package upgrade cap or we before we publish

thibault-martinez (approved):

PR #337, opened 2025-06-25
chore(indexer): Expose indexer 3030 port in docker compose

msarcev:
Add missing port mapping in docker compose


msarcev:
Already done nvm

PR #361, opened 2025-06-30
fix(dapp): Use correct nft id to update avatar

marc2332:
- use actual name id to update the avatar
- adds support to update avatar in subnames
- adds missing query invalidation after setting the avatar


KeitoTadashi (approved):
LGTM :rocket:

msarcev (approved):

Issue #250, opened 2025-06-03
Add action to renew name


begonaalvarezd:
Depends on https://github.com/iotaledger/iota-names/issues/247 & https://github.com/iotaledger/iota-names/issues/248, though it can be worked on the logic to renew and allowance criteria

- To all names and subnames, we should render a button to renew
- Renew is allowed if:
    - not expired
    - or expired:
        - names: if the grace period didnt expire yet (hardcoded value https://github.com/iotaledger/iota-names/blob/a0c665eabac80c504049a4c42e85c0ffc2f015d7/packages/iota-names/sources/constants.move#L19)
        - subnames: if parent is not expired
    - and, if subname, has renewal enabled
- Add logic to renew using the same expiration as parent

PR #368, opened 2025-07-01
chore(dapp): Simplify renewal updates creation

marc2332:


dapp/src/components/dialogs/RenewNameDialog.tsx
```diff
@@ -141,9 +152,10 @@ export function RenewNameDialog({ open, setOpen, name }: RenewDialogProps) {
         setOpen(false);
     };
 
+    const canRenew = nameRecord && updates.length > 0;
```
KeitoTadashi, line 151: not sure if we are following the right logic here
https://github.com/iotaledger/iota-names/issues/250#issue-3114885679

for tld: can't be renew if creation_timestamps - expiration_timestamp > 5 years
for sld: current logic is correct

maybe im missing something here but what 'updates.lenght > 0' means?

marc2332: canRenew means that the name record is loaded and that even there are no updates pushed, and therefore nothing to renew

I also added a new "wantsToRenew" which basicall means that there is a value in the years input

KeitoTadashi (approved):
LGTM

KeitoTadashi (approved):

KeitoTadashi (approved):

dapp/src/components/dialogs/CreateSubnameDialog.tsx
```diff
@@ -59,13 +59,13 @@ export function CreateSubnameDialog({ name, open, setOpen }: CreateSubnameProps)
     // Only join names if there user has written anything
     const fullSubdomainName = newSubdomainName.trim() ? newSubdomainName + '.' + name : null;
     const isAvailable = fullSubdomainName
-        ? getNameObject([], subdomainsOwned ?? [], fullSubdomainName) === null
+        ? getNameObject(subdomainsOwned ?? [], fullSubdomainName) === null
```
evavirseda: shouldn't it just be `subnamesOwned`? :thinking: and `newSubname` instead of `newSubdomainName`??

dapp/src/components/dialogs/RenewNameDialog.tsx
```diff
@@ -45,66 +98,19 @@ export function RenewNameDialog({ open, setOpen, name }: RenewDialogProps) {
...
-        ? getNamePermissions(nameRecord.nameRecord)
-        : null;
 
     // Editable values
     const [editRenewYears, setEditRenewYears] = useState<number>();
 
-    const { data: domainsOwned } = useRegistrationNfts('domain');
-    const { data: subdomainsOwned } = useRegistrationNfts('subdomain');
+    const { data: ownedNames } = useRegistrationNfts('domain');
+    const { data: ownedSubnames } = useRegistrationNfts('subdomain');
```
evavirseda: and should we also change the type 
```RegistrationNftType = 'domain' | 'subdomain'```
in `dapp/src/hooks/useRegistrationNfts.ts` to use name and subname instead of domain and subdomain?? :pray:

evavirseda (changes requested):
nits :pray:

msarcev:
image

Doesn't work for me, the button doesn't activate

evavirseda (approved):
works!

Issue #175, opened 2025-05-01
Improve packages coverage


thibault-martinez:
Using a debug iota binary:
`iota move test --coverage && iota move coverage summary`
`iota move coverage summary --summarize-functions`
`iota move coverage source --module <module_name>`

### `auction` 100%

- [x] `auction` 100%

### `coupons` 100 %

- [x] `range` 100.00%
- [x] `rules` 100.00%
- [x] `coupon` 100.00%
- [x] `coupons`100.00%
- [x] `coupon_house` 100.00%

### `iota-names` 99.16

- [ ] `name` 98.20% (@thibault-martinez)
- [x] `constants` 60.00% (@Thoralf-M) (doesn't really make sense to assert consts, some consts are used by the subdomain package, but that doesn't show up in the coverage for the main package)
- [x] `iota_names_registration` 100.00% (@Alex6323)
- [x] `subdomain_registration` 100.00%
- [x] `name_record` 100.00%
- [x] `iota_names` 90.26% (@DaughterOfMars) init function can't be tested
- [x] `registry` 100% (@Thoralf-M)
- [x] `core_config` 100.00%
- [x] `admin` 100.00%
- [x] `controller` 100.00% (@Alex6323)
- [x] `pricing_config` 100% (@DaughterOfMars)
- [x] `payment` 100% (@thibault-martinez)
- [x] `deny_list` 100%

### `payments` 100%

- [x] `payments` 100% (@Thoralf-M)

### `subnames` 99.63%

- [x] `config` 100.00% (@Alex6323)
- [x] `subnames` 99.53% (@DaughterOfMars)

### `temp-subname-proxy` 0% (just a wrapper around subnames)

- [ ] `subname_proxy` 0%

PR #373, opened 2025-07-02
feat(packages): improve test coverage

Thoralf-M:
From
+-------------------------+
| % Move Coverage: 95.93  |
+-------------------------+
to
+-------------------------+
| % Move Coverage: 98.47  |
+-------------------------+

https://github.com/iotaledger/iota-names/issues/175


DaughterOfMars:
What prevented 100%?

Thoralf-M:
> What prevented 100%?

I remember there was one section I couldn't find a test to cover last time, not sure if there is more, also for constants for example I don't see a point in adding a test that just asserts a const
Coverage can most probably still be increased though, just didn't want to spend more time on it now

DaughterOfMars (approved):

thibault-martinez (approved):

PR #376, opened 2025-07-02
fix(dapp): Not show `0` when there are no subnames

marc2332:
Was caused by the `subdomains?.length`


dapp/src/app/(protected)/my-names/page.tsx
```diff
@@ -101,54 +101,52 @@ export default function MyNamesPage(): JSX.Element {
             <div className="pt-md">
                 <Title title="My subnames" />
             </div>
-            {subdomains?.length && (
```
VmMad, line 104: Wouldn't it be better to do `subdomains?.length > 0` so that the div inside doesn't render if there are no subdomains?

marc2332: Maybe, but we are doing this exact thing with the names section from above, so I just went with it, when we actually focus on the UI we can definitely improve this, its just that I didnt want a 0 to show up lol

VmMad (approved):

KeitoTadashi (approved):
:medal_military:

PR #377, opened 2025-07-02
fix(dapp): Add grace period

### LINKED ISSUE #372: Add grace period when renewing a root name

KeitoTadashi:
fixes #372 

https://github.com/iotaledger/iota-names/issues/250


dapp/src/lib/utils/names.ts
```diff
@@ -1,14 +1,17 @@
...
+import { GRACE_PERIOD_MS, isSubName, NameRecord } from '@iota/iota-names-sdk';
 
 import { RegistrationNft } from '@/hooks';
 
 export function isNameRecordExpired(nameRecord: NameRecord) {
     return nameRecord.expirationTimestampMs < Date.now();
 }
 
+export function isNameRecordExpiredWithGracePeriod(nameRecord: NameRecord) {
+    return nameRecord.expirationTimestampMs < Date.now() + GRACE_PERIOD_MS;
```
msarcev: 1. I think `isGracePeriodExpired` is a clearer name.
2. This logic looks incorrect to me 🤔 `return nameRecord.expirationTimestampMs + GRACE_PERIOD_MS < Date.now();` makes more sense and it's also clearer.
Wdyt ?

msarcev (changes requested):

dapp/src/lib/utils/names.ts
marc2332: Yes, I agree with mario too

KeitoTadashi: agree

msarcev (approved):

marc2332 (changes requested):
What about the call of `isNameRecordExpired` in `CreateSubnameDialog.tsx` ? Is that not needed there as well? Only in renew?

marc2332 (approved):

PR #378, opened 2025-07-02
feat(indexer): skip syncing checkpoints before SC deployment

### LINKED ISSUE #375: Find and use IOTA-Names deployment checkpoint as start checkpoint for indexer
Instead of starting to sync from checkpoint 0, find the checkpoint where the iota-names package was deployed and only sync from this checkpoint, by setting it for the progress_store the first time the indexer is started
Without this, the indexer potentially needs to sync many days of data for nothing

Thoralf-M:
Skip syncing checkpoints before IOTA-Names SC deployment by first fetching the checkpoint in which the SC was published and setting that in the progress_store.

Fixes https://github.com/iotaledger/iota-names/issues/375

Tested by running `docker compose -f indexer/docker/docker-compose.yml up`, it skipped all the earlier checkpoints as expected


DaughterOfMars (approved):

indexer/docker/docker-compose.yml
```diff
@@ -40,7 +40,12 @@ services:
...
     command:
       - "start"
-      - "--node-url=http://host.docker.internal:9000"
+      # Localnet values:
+      # - "--node-url=http://host.docker.internal:9000"
+      # - "--checkpoint-url=http://host.docker.internal:9000"
+      # Devnet values:
+      - "--node-url=https://api.devnet.iota.cafe"
+      - "--checkpoint-url=https://checkpoints.devnet.iota.cafe"
       - "--prometheus-url=http://prometheus:9090"
```
thibault-martinez: Can you put this one before the network dependent ones please? Someone may inadvertently comment it thinking it's part of the devnet stuff

Thoralf-M: Sure, https://github.com/iotaledger/iota-names/pull/378/commits/3500097422b4a77b7105a20349979962ed07fe69

indexer/src/worker.rs
```diff
@@ -65,14 +71,14 @@ pub(crate) async fn run_iota_names_reader(
     );
     executor.register(worker_pool).await?;
 
-    info!("Connecting to node at {node_url}");
+    info!("Connecting to {checkpoint_url} to sync checkpoints");
     // Localnet does not support remote store, so we use the REST API
-    if node_url.contains("localhost") || node_url.contains("host.docker.internal") {
+    if checkpoint_url.contains("localhost") || checkpoint_url.contains("host.docker.internal") {
```
thibault-martinez, line 76: Why was the node URL not sufficient anymore?

Thoralf-M: Because the checkpoints URL doesn't provide the JSON API we need to get the checkpoint when the package was published

thibault-martinez: I'll approve but I'm still a bit confused. We were previously using the node URL for the historical URL and live URL. You are now also using the node URL for the client to get the checkpoint or the package. Why introduce the checkpoint URL?

thibault-martinez (approved):

indexer/src/worker.rs
Thoralf-M: For localnet we can use the same URL, as we have the REST API and the JSON API exposed
However for the public network endpoints it's different, there we have either JSON API or the checkpoints API for the IndexerExecutor, but not both

Thoralf-M: So for devnet this PR renames node_url to checkpoint_url and adds a node_url with an actual node

thibault-martinez: Alright, thanks

PR #383, opened 2025-07-03
feat(names-ts-sdk): Use the same constants format in both sdk and scripts

marc2332:
Adds missing values and moves it from one place to another, this way its way simpler to copy and paste the values directly to constants.ts

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ iota.log.*
 .coverage_map.mvcov
 /scripts/tx/*
 /scripts/package-info/localnet.json
+/scripts/package-info/sdk.json
 !/scripts/tx/.gitkeep
 localnet.json
 
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@ iota.log.*
 .coverage_map.mvcov
 /scripts/tx/*
 /scripts/package-info/localnet.json
-/scripts/package-info/sdk.json
 !/scripts/tx/.gitkeep
 localnet.json
 
--- a/scripts/init/publish.ts
+++ b/scripts/init/publish.ts
@@ -60,26 +60,5 @@ export const publishPackages = async (network: string, isCiJob = false, configPa
         JSON.stringify(results, null, 2),
     );
     console.log('******* Packages published successfully *******');
-    const data = results as PackageInfo;
-
-    // Export the constants based on the SDK's format so SDK can be easily tested.
-    writeFileSync(
-        path.resolve(path.resolve(__dirname, '../'), 'constants.sdk.json'),
-        JSON.stringify(
-            {
-                iotaNamesPackageId: {
-                    latest: data.IotaNames.packageId,
-                    v1: data.IotaNames.packageId,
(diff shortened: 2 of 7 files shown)


msarcev (approved):

.gitignore
```diff
@@ -8,6 +8,7 @@ iota.log.*
 .coverage_map.mvcov
 /scripts/tx/*
 /scripts/package-info/localnet.json
+/scripts/package-info/sdk.json
```
thibault-martinez: What's that?

scripts/init/publish.ts
```diff
@@ -60,26 +60,5 @@ export const publishPackages = async (network: string, isCiJob = false, configPa
         JSON.stringify(results, null, 2),
     );
     console.log('******* Packages published successfully *******');
-    const data = results as PackageInfo;
-
-    // Export the constants based on the SDK's format so SDK can be easily tested.
-    writeFileSync(
```
thibault-martinez, line 66: So this is not called anymore on publishing?

.gitignore
marc2332: its basically localnet.json but with the format that the TS SDK works with

scripts/init/publish.ts
marc2332: It is, just no here, because here we dont have all the information that we need, that's why I put just after the `[network].json` gets generated, there we have the missing bits

thibault-martinez:
conflicts

marc2332:
> conflicts

Done

scripts/package-info/constants.ts
```diff
@@ -38,4 +38,27 @@ export const writePackageInfo = (network: string, packageInfo: PackageInfo) => {
...
+            {
+                auctionPackageId: packageInfo.auctionPackageId,
+                auctionHouseObjectId: packageInfo.auctionHouseObjectId,
+                packageId: packageInfo.packageId,
+                iotaNamesObjectId: packageInfo.iotaNamesObjectId,
+                subnamesPackageId: packageInfo.subnamesPackageId,
+                tempSubnameProxyPackageId: packageInfo.tempSubnameProxyPackageId,
+                payments: {
+                    packageId: packageInfo.paymentsPackageId,
+                },
```
thibault-martinez: Can we make this not an object, like the other package IDs?

marc2332: Yes, but if I am gonna change this I might as well just remove the sdk file generation and use the same format as the network jsons

thibault-martinez: Sounds like a great idea

marc2332: > Yes, but if I am gonna change this I might as well just remove the sdk file generation and use the same format as the network jsons

Done, I used the same type and data. Some fields will obviously not be used but its not a big deal. Now its as easy as copying from the `[network].json` to `constants.ts`

thibault-martinez (approved):

msarcev (approved):

PR #397, opened 2025-07-07
feat(dapp): validate renew years

### LINKED ISSUE #363: Show max years/time when renewing names/subnames
Needs https://github.com/iotaledger/iota-names/pull/263 first

`RenewNameDialog.tsx`

### LINKED ISSUE #382: Validate maximum years for renew name from SC and render a proper error
Image

SC only allow to renew name a max of 5 years.

KeitoTadashi:
:warning: need first #435 

fixes #382 
fixes #363


evavirseda (changes requested):
something is not working i think :pray: 
image

dapp/src/components/dialogs/RenewNameDialog.tsx
```diff
@@ -140,7 +140,7 @@ export function RenewNameDialog({ open, setOpen, name }: RenewDialogProps) {
     const handleCancelRenewName = () => {
         setOpen(false);
     };
-
+    const exceedYears = updateNameError && updateNameError?.message.includes('9223373020403073037');
```
marc2332: We should definitely not hardcode the error codes in our code, I would rather do something else like a manual check that the amount of years is lower then the max

marc2332 (changes requested):

evavirseda (changes requested):
Still seeing some weird behavior :see_no_evil:  the errors aren’t consistent, even though i’m supposed to exceed the max number of years in both cases

image

image

dapp/src/components/dialogs/RenewNameDialog.tsx
```diff
@@ -147,11 +147,29 @@ export function RenewNameDialog({ open, setOpen, name }: RenewDialogProps) {
...
+        }
+        return 0;
+    }
+
+    const isRenewable =
+        nameRecord?.nameRecord?.expirationTimestampMs && editRenewYears
+            ? nameRecord.nameRecord.expirationTimestampMs +
+                  editRenewYears * 365 * 24 * 60 * 60 * 1000 <
+              Date.now() + 6 * 365 * 24 * 60 * 60 * 1000
+            : true;
```
marc2332: All this logic should not be in this component, this should be a util function or something, also, hardcoding the `5` years here is a no-go, we should be able to fetch this from the core config or something

dapp/src/components/dialogs/RenewNameDialog.tsx
```diff
@@ -165,20 +183,35 @@ export function RenewNameDialog({ open, setOpen, name }: RenewDialogProps) {
...
+                                        setRenewError(null);
+                                        if (val > 5) {
+                                            setRenewError(
+                                                'You cannot renew for more than 5 years.',
+                                            );
+                                        } else if (val < 0) {
+                                            setRenewError('Input a positive number.');
+                                        } else {
+                                            setEditRenewYears(isNaN(val) ? 0 : val);
+                                        }
```
marc2332: same, hardcoding the 5 here is a no-go, also, I dont see why we need an error state, we can just derive this from the "editRenewYears" state

dapp/src/components/dialogs/RenewNameDialog.tsx
```diff
@@ -101,7 +101,7 @@ export function RenewNameDialog({ open, setOpen, name }: RenewDialogProps) {
 
     // Editable values
     const [editRenewYears, setEditRenewYears] = useState<number>();
-
+    const [renewError, setRenewError] = useState<string | null>(null);
```
marc2332:
```suggestion
```

marc2332 (changes requested):

sdk/src/constants.ts
```diff
@@ -5,7 +5,7 @@
 export const MAX_U64 = BigInt('18446744073709551615');
 export const MIN_LABEL_SIZE = 3;
 export const GRACE_PERIOD_MS = 30 * 24 * 60 * 60 * 1000;
-
+export const MAX_RENEWAL_YEARS = 5;
```
evavirseda: i don't think this is what marc suggested [here](https://github.com/iotaledger/iota-names/pull/397#discussion_r2197164459) 🙈 he proposed adding it to the core config, we can do it in another issue if you prefer 🙏

evavirseda (changes requested):

sdk/src/constants.ts
KeitoTadashi: @marc2332 is that even possible?

marc2332: Yes, I'll do it myself in a separate pr

evavirseda (dismissed):

sdk/src/constants.ts
marc2332: my pr has been merged

marc2332 (changes requested):
I will push some changes to this PR.

marc2332 (approved):

evavirseda (commented):
I think this PR is no longer necessary 🙏  since the reskin of the renew popup replaced the input with a select, which already handles the limitation on the number of years that can be renewed 🙈 also, the `useCoreConfig` hook already exists in develop

evavirseda:
No longer needed :pray:

PR #400, opened 2025-07-08
feat(dapp): add balance validation to auctions

### LINKED ISSUE #365: Improve balance validation for AuctionBid and PurchaseName dialogs
We should add a balance validation when bidding an auction, in order to prevent failed TX or showing smart contracts errors

https://github.com/user-attachments/assets/3e23b103-ba87-4e19-9911-aaa25698da82

EDIT:

Let's add a `useBalanceValidation` hook that will be used in the two dialogs. This hook will:
- take the built transaction, when available
- do a "dryRun" to have the gas info
- pessimistically compute the hasBalance and hasEnoughGas replacing logic in L97-104 of PurchaseNameDialog
- return the validation info needed for the components

The error about missing balance should be sanitized and user friendly.

KeitoTadashi:
fixes #365


msarcev:
Let's add balance validation with `zod` 🙏🏻

dapp/src/components/dialogs/PurchaseNameDialog.tsx
```diff
@@ -145,12 +145,44 @@ export function PurchaseNameDialog({ name, open, setOpen, onPurchase }: Purchase
...
+                            {nameRecordError && (
+                                <div className="text-center text-red-400 text-sm">
+                                    {nameRecordError.message}
+                                </div>
+                            )}
+                            {hasEnoughGas && updateNameError && (
+                                <div className="text-center text-red-400 text-sm">
+                                    {updateNameError.message}
+                                </div>
+                            )}
```
VmMad: In these cases, the best option is to just use the InfoBox

VmMad (changes requested):
Nice progress. I'd add two more things:

1. Change bid and purchase inputs to be type NumericFormat, like we do in the dashboard to prevent weird inputs like 
image

2. Seems to be missing a not enough balance validation
image

VmMad (approved):

evavirseda (approved):

PR #430, opened 2025-07-11
Update purchase and auction dialog funcionality

### LINKED ISSUE #417: Add missing features to purchase & auction
depends on https://github.com/iotaledger/iota-names/pull/411

implement missing logic and styles for the following features according to the specifications in issue https://github.com/iotaledger/iota-names/issues/347
- expiration selector
- set name as default
- all fiat stuff

KeitoTadashi:
fixes #417


dapp/src/components/dialogs/PurchaseNameDialog.tsx
```diff
@@ -41,11 +44,22 @@ export function PurchaseNameDialog({ name, open, setOpen, onPurchase }: Purchase
     const queryClient = useQueryClient();
     const client = useIotaClient();
     const account = useCurrentAccount();
+
+    const [renewYears, setRenewYears] = useState<number>(1);
+    const [isDisplayName, setIsDisplayName] = useState<boolean>(false);
+
+    // const { data: subnamesOwned } = useRegistrationNfts('subname');
```
evavirseda: 🧹

dapp/src/components/dialogs/PurchaseNameDialog.tsx
```diff
@@ -142,8 +157,33 @@ export function PurchaseNameDialog({ name, open, setOpen, onPurchase }: Purchase
...
+                                <div className="flex flex-row gap-x-sm w-full p-md">
+                                    <Checkbox
+                                        isChecked={isDisplayName}
+                                        onCheckedChange={(e) => setIsDisplayName(e.target.checked)}
+                                        label="Set name as Display Name"
+                                    />
+                                </div>
+                            </Panel>
                             <div className="flex flex-row gap-x-sm w-full">
                                 <DisplayStats label="Registration Expires" value={expirationDate} />
```
evavirseda, line 191: i think the `expirationDate` should be updated after selecting a different expiration option 🙏

dapp/src/components/dialogs/PurchaseNameDialog.tsx
```diff
@@ -142,8 +157,33 @@ export function PurchaseNameDialog({ name, open, setOpen, onPurchase }: Purchase
                                     </span>
                                 </div>
                             </Panel>
+                            <div className="px-md py-sm border-t border-names-neutral-6">
+                                <Select
```
evavirseda, line 170: we shouldn't hardcode the values here, take them from the core config 🙏 you can take a look at the `RenewNameDialog` to see how it works

evavirseda: also, if we set it to 2 years and then go to the renew dialog, it should detect how many years are left, but right now, I can buy a name with 3 years and then renew it for 5 more 🙈  and i don’t think that’s supposed to happen 🙏

dapp/src/lib/utils/format/sanitizeIotaName.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+export function sanitizeIotaName(name: string) {
```
evavirseda: duplicated function, there’s already one in `/utils/format` called `denormalizeName` that does the same thing 🙏

evavirseda (changes requested):

dapp/src/lib/utils/getDefaultExpirationDate.ts
```diff
@@ -6,3 +6,7 @@ import { formatExpirationDate } from './format/formatExpirationDate';
 export function getDefaultExpirationDate(): string {
     return formatExpirationDate(new Date(Date.now() + 365 * 24 * 60 * 60 * 1000));
 }
+
+export function getDefaultExpirationDateWithRenewYears(renewYears: number): string {
```
evavirseda: mmm im not sure about this function, we don’t use it anywhere else, so i don’t think it’s necessary to move it to utils 🙏

evavirseda (changes requested):
tiny detail 🙏

evavirseda (approved):

dapp/src/components/dialogs/PurchaseNameDialog.tsx
```diff
@@ -14,21 +15,29 @@ import {
...
     GAS_BALANCE_TOO_LOW_ID,
     GAS_BUDGET_ERROR_MESSAGES,
     NOT_ENOUGH_BALANCE_ID,
 } from '@/lib/constants';
 import { formatNanosToIota } from '@/lib/utils';
+import { formatExpirationDate } from '@/lib/utils/format/formatExpirationDate';
 import { denormalizeName } from '@/lib/utils/format/formatNames';
-import { getDefaultExpirationDate } from '@/lib/utils/getDefaultExpirationDate';
+
+function getDefaultExpirationDateWithRenewYears(renewYears: number): string {
```
msarcev: This needs to go into a `util` somewhere and be something like:
```
function getTargetExpirationDate(renewYears: number): string {
    const today = new Date();
    return formatExpirationDate(new Date(today.setFullYear(today.getFullYear() + renewYears)));
}
```

msarcev (changes requested):

marc2332 (approved):

msarcev (approved):

Issue #314, opened 2025-06-23
Final cleanup before community testing


thibault-martinez:
- [x] @DaughterOfMars 
- [x] @Thoralf-M 
- [x] @Alex6323 
- [x] @thibault-martinez

DaughterOfMars:
Shall we close this?

thibault-martinez:
> Shall we close this?

I have one or two things I know I want to double check, I'll close it

PR #450, opened 2025-07-14
chore(*): clean up `iota-names` packages

Alex6323:
cleans up docs, tries to reduce some inconsistencies

Addresses #314


packages/auction/sources/auction.move
```diff
@@ -77,7 +77,11 @@ fun init(ctx: &mut TxContext) {
     });
 }
 
-/// Start an auction if it's not started yet; and make the first bid.
+/// Start an auction for a name if it's available and not being auctioned already.
+/// 
+/// ### NOTE
+/// 
+/// This method will also place the first bid.
```
DaughterOfMars:
```suggestion
/// Start an auction for a name by placing the first bid.
/// This will fail if the name is not available or already being auctioned.
```

packages/auction/sources/auction.move
```diff
@@ -121,13 +125,12 @@ public fun start_auction_and_place_bid(
...
 /// Bidders use this function to place a new bid.
 ///
-/// Panics
-/// Panics if `name` is invalid
-/// or there isn't an auction for `name`
-/// or `bid` is too low,
+/// ### Panics
+/// 
+/// Panics if `name` is invalid or there isn't an auction
+/// for `name` or `bid` is too low.
```
DaughterOfMars:
```suggestion
/// Panics if:
/// - the name is invalid
/// - there is no auction for the name
/// - the bid is too low
```

packages/coupons/sources/coupons.move
```diff
@@ -51,25 +47,35 @@ public fun add_fixed_coupon(
     self.save_coupon(hash, coupon::new_fixed(amount, rules));
 }
 
-// A function to remove a coupon from the system.
+// === Internal functions ===
+
+/// An internal function to create the `Coupons` struct.
```
DaughterOfMars: I'd rather remove all the "An internal function to ... " and just say what it does

packages/auction/sources/auction.move
Alex6323: I think this is missing a "Panics if:" or smth?

DaughterOfMars: I'll let you decide, just figured a list was better here

packages/coupons/sources/coupons.move
Alex6323: https://github.com/iotaledger/iota-names/pull/450/commits/1482ddf03b7a1982f57642f0871118cfb2361218

packages/iota-names/sources/constants.move
```diff
@@ -15,19 +15,17 @@ use std::string::String;
...
 /// The amount of milliseconds in a year.
 const YEAR_MS: u64 = 365 * 24 * 60 * 60 * 1000;
-/// 30 day Grace period in milliseconds.
+/// 30 day grace period in milliseconds.
 const GRACE_PERIOD_MS: u64 = 30 * 24 * 60 * 60 * 1000;
 
+// === Subname constants ===
+// These constants are the core of the subname functionality.
+// Even if we decide to change the subname module, these can
+// be re-used. They're added as metadata on `NameRecord`.
```
DaughterOfMars, line 28: Technically it was in the right place before, because the expiration is not metadata but a field

packages/iota-names/sources/controller.move
```diff
@@ -47,7 +47,7 @@ public fun set_target_address(
     registry.set_target_address(name, new_target);
 }
 
-/// Set the reverse lookup address for the name
+/// Set the reverse lookup address for the name.
```
DaughterOfMars:
```suggestion
/// Sets the reverse lookup address for the name.
```

packages/iota-names/sources/name_registration.move
```diff
@@ -98,7 +91,7 @@ public fun expiration_timestamp_ms(self: &NameRegistration): u64 {
     self.expiration_timestamp_ms
 }
 
-// get a read-only `uid` field of `NameRegistration`.
+// Get a read-only `uid` field of `NameRegistration`.
```
DaughterOfMars:
```suggestion
// Returns a read-only `uid` field of `NameRegistration`.
```

packages/iota-names/sources/registry.move
```diff
@@ -185,13 +185,13 @@ public fun add_leaf_record(
 ) {
     assert!(name.is_subname(), EInvalidDepth);
 
-    // get the parent of the name
+    // Get the parent of the name.
```
DaughterOfMars:
```suggestion
    // Returns the parent of the name.
```

packages/iota-names/sources/constants.move
Alex6323: ah, okay, will change it back then. Thanks

packages/subnames/sources/config.move
```diff
@@ -45,7 +47,7 @@ public fun default(): SubnameConfig {
     }
 }
 
-// Generates a custom config for Subnames.
+/// Generates a custom config for Subnames.
```
Thoralf-M, line 50: Generates sounds weird here

Alex6323: What do you prefer?

DaughterOfMars: I think it's fine

DaughterOfMars (approved):

packages/subnames/sources/config.move
Thoralf-M: In the other places we have Creates, but yeah probably not so important

Thoralf-M (approved):

packages/iota-names/sources/constants.move
Alex6323: https://github.com/iotaledger/iota-names/pull/450/commits/3709e915147e1ae29cc4223a927689ca6c6a8f8c

PR #454, opened 2025-07-15
chore: Add READMEs

DaughterOfMars:
## Description

Adds README files for the various packages with some basic info.

### DIFF
--- a/packages/auction/README.md
+++ b/packages/auction/README.md
@@ -4,7 +4,7 @@ This package is for the initial IOTA-Names launch. It provides an alternative pu
 
 ## Process
 
-In order to purchase a name via this method, one must begin by starting a new auction for a particular name. The `start_auction_and_place_bid` function accepts an initial bid, which must be at minimum the base price of the name. This will create the NFT and register it within IOTA-Names, though it will be initially owned by the Auction House. 
+In order to purchase a name via this method, one must begin by starting a new auction for a particular name. The `start_auction_and_place_bid` function accepts an initial bid, which must be at minimum the base price of the name. This will create the NFT and register it within IOTA-Names, though it will be initially owned by the Auction House.
 
 Until the end of the auction period, which is defined by `AUCTION_BIDDING_PERIOD_MS`, anyone may place additional bids on the name via the `place_bid` function. Each bid must be greater than the last by at least `AUCTION_MIN_OVERBID_VALUE_IOTA`. When a new bid is successfully placed, the previous bid will be returned to the owner.
 
--- a/packages/auction/README.md
+++ b/packages/auction/README.md
@@ -8,6 +8,6 @@ In order to purchase a name via this method, one must begin by starting a new au
 
 Until the end of the auction period, which is defined by `AUCTION_BIDDING_PERIOD_MS`, anyone may place additional bids on the name via the `place_bid` function. Each bid must be greater than the last by at least `AUCTION_MIN_OVERBID_VALUE_IOTA`. When a new bid is successfully placed, the previous bid will be returned to the owner.
 
-Once the auction period is over, there is an additional quiet period to allow for last-minute bids. When some time, defined by `AUCTION_MIN_QUIET_PERIOD_MS`, has passed without any additional bids, the auction will be locked.
+Once the auction period is over, there may be an additional quiet period to allow for last-minute bids. When a bid is made near the end of the period, a small amount of additional time is added, defined by `AUCTION_MIN_QUIET_PERIOD_MS`. Once this full time has passed without any additional bids, the auction will be locked.
 
 Finally, the winner may call the `claim` function to accept their NFT.
--- a/packages/auction/README.md
+++ b/packages/auction/README.md
@@ -1,1 +1,13 @@
-Auctions package for the initial IOTA-Names launch to ensure fairer registration.
+# Auctions
+
+This package is for the initial IOTA-Names launch. It provides an alternative purchase path via auctions in order to ensure fairer registration.
+
+## Process
+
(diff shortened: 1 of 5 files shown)


packages/auction/README.md
```diff
@@ -1 +1,13 @@
...
+
+This package is for the initial IOTA-Names launch. It provides an alternative purchase path via auctions in order to ensure fairer registration.
+
+## Process
+
+In order to purchase a name via this method, one must begin by starting a new auction for a particular name. The `start_auction_and_place_bid` function accepts an initial bid, which must be at minimum the base price of the name. This will create the NFT and register it within IOTA-Names, though it will be initially owned by the Auction House. 
+
+Until the end of the auction period, which is defined by `AUCTION_BIDDING_PERIOD_MS`, anyone may place additional bids on the name via the `place_bid` function. Each bid must be greater than the last by at least `AUCTION_MIN_OVERBID_VALUE_IOTA`. When a new bid is successfully placed, the previous bid will be returned to the owner.
+
+Once the auction period is over, there is an additional quiet period to allow for last-minute bids. When some time, defined by `AUCTION_MIN_QUIET_PERIOD_MS`, has passed without any additional bids, the auction will be locked.
```
Thoralf-M: AUCTION_MIN_QUIET_PERIOD_MS is only added if the auction end time is  less than this, it's not always added additional to the normal auction period

DaughterOfMars: Yeah I tried to say that

packages/subnames/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+Node subnames can only be burned once they have expired.
+
+### Leaf Subnames
+
+Leaf subnames do not have an associated NFT, they only exist as a record in the IOTA-Names registry. This means they have several limitations compared to nodes:
+
+- They cannot have subnames
+- They cannot be individually transferred
+- Their expiration is tied to their parent
```
Thoralf-M, line 28: Not sure if it fits here, but should we mention here that there is no API to know which leafs one owns?

packages/auction/README.md
Thoralf-M: But `Once the auction period is over, there is an additional quiet period to allow for last-minute bids.` to me is clearly auction period + quite period
Should be more like `If a bid is placed within the last AUCTION_MIN_QUIET_PERIOD_MS of the auction duration, the auction duration gets extended by AUCTION_MIN_QUIET_PERIOD_MS`

packages/subnames/README.md
DaughterOfMars: Yeah I thought about it but I don't think it fits

packages/subnames/README.md
```diff
@@ -0,0 +1,30 @@
+# Subnames
+
+This packages provides functions to create and manage subnames. Subnames are free to create as long as the parent name is owned, and any number of subnames can be created.
```
Alex6323:
```suggestion
This package provides functions to create and manage subnames. Subnames are free to create as long as the parent name is owned, and any number of subnames can be created.
```

packages/auction/README.md
DaughterOfMars: Tried to rephrase it

packages/subnames/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Types of Subname
+
+Subnames can be one of two different types: leaf or node. Each has different use-cases and limitations.
+
+### Node Subnames
+
+Node subnames have an associated NFT (`SubnameRegistration`), an expiration date and time, and a set of rules which govern their usage. The NFT defines the ownership of the subname, and can be used to manage it, transfer it, and burn it.
+
+When creating a node subname, an expiration date and time must be provided. This must be after than the minimum time defined by `MINIMUM_SUBNAME_DURATION` and before the expiration of the parent.
```
Alex6323:
```suggestion
When creating a node subname, an expiration date and time must be provided. This must be after the minimum time defined by `MINIMUM_SUBNAME_DURATION` and before the expiration of the parent.
```

packages/subnames/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+### Leaf Subnames
+
+Leaf subnames do not have an associated NFT, they only exist as a record in the IOTA-Names registry. This means they have several limitations compared to nodes:
+
+- They cannot have subnames
+- They cannot be individually transferred
+- Their expiration is tied to their parent
+
+Leaf subnames can also be removed at will, regardless of the parent expiration.
```
Alex6323, line 30:
```suggestion
Leaf subnames can also be removed at will, regardless of the parent expiration.

#### Recommendation

For most use-cases a Node Subname is the right choice as it provides the owner with full flexibility to manage it. Therefore, it is recommended to use Leaf Subnames only in very special scenarios where one parent needs to control a large amount of subname siblings and also can easily keep track of them off-chain (e.g. employee1@company, employee2@company, ...)
```

Should we add a recommendation when to use what? I added an example of what we *could* write here. Feel free to replace it with whatever you prefer.

DaughterOfMars: I don't think so, personally

Alex6323: fair enough :+1: Are there other opinions? @thibault-martinez @Thoralf-M 

Personally, I think, some guidance here is helpful to make clear that Leafs aren't what users want in 99% of the time.

Alex6323 (approved):

packages/subnames/README.md
Thoralf-M: I also thought about it, but it's probably better to mention this in the other docs, this README is specific to just what the package contains and not really about how to use it

Thoralf-M (approved):

PR #469, opened 2025-07-16
fix(scripts): adapt to keystore changes

### LINKED ISSUE #463: Update to latest keystore changes
Since the format changed, the script here needs to be adapated
```
iota-names/scripts/utils/utils.ts:175
    for (const priv of keystore) {
                       ^
TypeError: keystore is not iterable
    at getSigner (/home/t/Desktop/iota-names/scripts/utils/utils.ts:175:24)
    at signAndExecute (/home/t/Desktop/iota-names/scripts/utils/utils.ts:196:29)
    at publishPackages (/home/t/Desktop/iota-names/scripts/init/publish.ts:41:45)
    at init (/home/t/Desktop/iota-names/scripts/init/init.ts:37:44)
    at Object.<anonymous> (/home/t/Desktop/iota-names/scripts/init/init.ts:102:5)
    at Module._compile (node:internal/modules/cjs/loader:1529:14)
    at Module.m._compile (/home/t/Desktop/iota-names/node_modules/.pnpm/ts-node@10.9.2_@types+node@20.17.27_typescript@5.8.3/node_modules/ts-node/src/index.ts:1618:23)
    at Module._extensions..js (node:internal/modules/cjs/loader:1613:10)
    at Object.require.extensions.<computed> [as .ts] (/home/t/Desktop/iota-names/node_modules/.pnpm/ts-node@10.9.2_@types+node@20.17.27_typescript@5.8.3/node_modules/ts-node/src/index.ts:1621:12)
    at Module.load (node:internal/modules/cjs/loader:1275:32)
```

Thoralf-M:
Fixes https://github.com/iotaledger/iota-names/issues/463

### DIFF
--- a/scripts/utils/utils.ts
+++ b/scripts/utils/utils.ts
@@ -172,13 +172,12 @@ export const getSigner = () => {
         readFileSync(path.join(homedir(), '.iota', 'iota_config', 'iota.keystore'), 'utf8'),
     );
 
-    for (const priv of keystore) {
-        const keypair = decodeIotaPrivateKey(priv);
-
-        const pair = Ed25519Keypair.fromSecretKey(keypair.secretKey);
-        if (pair.getPublicKey().toIotaAddress() === sender) {
-            return pair;
+    for (const entry of keystore.keys) {
+        if (entry.key.type != 'key_pair' || entry.address != sender) {
+            continue;
         }
+        const keypair = decodeIotaPrivateKey(entry.key.value);
+        return Ed25519Keypair.fromSecretKey(keypair.secretKey);
     }
 
     throw new Error(`keypair not found for sender: ${sender}`);
--- a/scripts/utils/utils.ts
+++ b/scripts/utils/utils.ts
@@ -182,6 +182,7 @@ export const getSigner = () => {
                 return pair;
             }
         }
+        throw new Error(`keypair not found for sender: ${sender}`);
     }
 
     // New format
--- a/scripts/utils/utils.ts
+++ b/scripts/utils/utils.ts
@@ -175,9 +175,6 @@ export const getSigner = () => {
     // Support for old format, can be removed once https://github.com/iotaledger/iota/pull/7704 is in production
     if (Array.isArray(keystore)) {
         for (const priv of keystore) {
-            if (typeof priv !== 'string') {
-                break;
-            }
             const keypair = decodeIotaPrivateKey(priv);
 
             const pair = Ed25519Keypair.fromSecretKey(keypair.secretKey);
--- a/scripts/utils/utils.ts
+++ b/scripts/utils/utils.ts
@@ -172,6 +172,22 @@ export const getSigner = () => {
         readFileSync(path.join(homedir(), '.iota', 'iota_config', 'iota.keystore'), 'utf8'),
     );
 
+    // Support for old format, can be removed once https://github.com/iotaledger/iota/pull/7704 is in production
+    if (Array.isArray(keystore)) {
+        for (const priv of keystore) {
+            if (typeof priv !== 'string') {
+                break;
+            }
+            const keypair = decodeIotaPrivateKey(priv);
+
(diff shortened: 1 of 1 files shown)


Thoralf-M:
~CI fails as long as there is no new release with the keystore changes~
Changed it to work with both formats

DaughterOfMars (approved):

scripts/utils/utils.ts
```diff
@@ -172,13 +172,28 @@ export const getSigner = () => {
         readFileSync(path.join(homedir(), '.iota', 'iota_config', 'iota.keystore'), 'utf8'),
     );
 
-    for (const priv of keystore) {
-        const keypair = decodeIotaPrivateKey(priv);
+    // Support for old format, can be removed once https://github.com/iotaledger/iota/pull/7704 is in production
+    if (Array.isArray(keystore)) {
+        for (const priv of keystore) {
+            if (typeof priv !== 'string') {
+                break;
```
Alex6323: feels like this should be a `continue` instead of a `break` tbh :thinking:

scripts/utils/utils.ts
```diff
@@ -172,13 +172,28 @@ export const getSigner = () => {
...
-        const pair = Ed25519Keypair.fromSecretKey(keypair.secretKey);
-        if (pair.getPublicKey().toIotaAddress() === sender) {
-            return pair;
+    // New format
+    for (const entry of keystore.keys) {
+        if (entry.key.type !== 'key_pair' || entry.address !== sender) {
+            continue;
         }
+        const keypair = decodeIotaPrivateKey(entry.key.value);
+        return Ed25519Keypair.fromSecretKey(keypair.secretKey);
```
Alex6323, line 194: could this go into an `else` block? (it seems like for the new format `keystore` is not an array)

scripts/utils/utils.ts
Thoralf-M: The file would be broken if it's not a string, I just removed this part

scripts/utils/utils.ts
Thoralf-M: Sure, but that would just make it more nested

scripts/utils/utils.ts
Thoralf-M: https://github.com/iotaledger/iota-names/pull/469/commits/a59c3f4706abfc81736e1d596c1dc7bfc27c5f95

scripts/utils/utils.ts
Alex6323: but the lines 188...194 (for loop) shouldn't run if the `keystore` is an array, but it didn't return a pair, no? It should then throw the error of line 196 immediatedly?

Thoralf-M: Tbh we don't have to consider corrupted keystore files here, if it's corrupted it would fail at some point, tiny bit earlier or later doesn't matter

scripts/utils/utils.ts
Alex6323: ah, that was the only option, that the whole file is broken then :+1: (but yeah, I think a file validity check - if necessary - should be done when the keystore is loaded, not here)

scripts/utils/utils.ts
Alex6323: can we then add a comment, that provides this context, why lines 188...194 don't run unnecessarily if `keystore` is an array?

Thoralf-M: Happy with that? 
https://github.com/iotaledger/iota-names/pull/469/commits/1581915cd63ba19f82b3cdb1e5dc7612cbf853b3

Alex6323: I still prefer `else` because now you have repeated code which is worse than nested code IMHO, but sure, why not!

Alex6323 (approved):

scripts/utils/utils.ts
Thoralf-M: Can be removed in a few weeks anyways

PR #483, opened 2025-07-18
feat: Add `/auctions` API endpoint

### LINKED ISSUE #461: Add endpoint to get all auctions
+ support for pagination

DaughterOfMars:
## Description

Add an endpoint to fetch auctions which are sorted and paged.

#### Query params
- `page` (number, default: 1)
- `pageSize` (number, default: 20)
- `sortBy` (string, default: “name”) - “name” or “bid”
- `sortOrder` (string, default: “asc”) - Sort direction: “asc” or “desc”
- `search` (string, optional) - Substring to search in the names

Closes #461


indexer/src/db/mod.rs
Alex6323, line 1: `SortOrder` and `AuctionSortBy` are only ever used in a query context, right? So maybe they can be moved into the `queries` module?

DaughterOfMars: Those are DB queries though, so it's not exactly the same. I did consider breaking the api up into a folder but I didn't feel like it was worth it yet.

indexer/src/db/queries.rs
```diff
@@ -92,6 +93,32 @@ pub fn get_names_for_bidder_address(
...
+) -> Result<Vec<String>> {
+    let mut query = names::table
+        .inner_join(name_bids::table)
+        .select(names::name)
+        .limit(page_size as _)
+        .offset((page.unwrap_or_default() * page_size) as _)
+        .into_boxed();
+    if let Some(search) = search {
+        query = query.filter(names::name.like(format!("%{search}%")))
+    }
```
Thoralf-M, line 106: So if the first page contains nothing that matches the search substring filter, it would return an empty vec, but next page?
Shouldn't the filter be applied before the page size limit?

DaughterOfMars: I don't think the order matters, does it?

Thoralf-M: What I mean is the following, correct me if I'm wrong but I think now it works like this:
query contains `search: Some("name".into())`
table names are like this:
000.iota
001.iota
002.iota
003.iota
my-name.iota
004.iota
now if the page_size is just 2, it would just get 000.iota, 001.iota, then apply the search filter and return an empty vec
So you query the next page, there the same 002.iota, 003.iota -> empty vec
Only the third page would contain your my-name.iota entry

Instead `my-name.iota` should be returned in the first page already and the other entries should be skipped, this also means the page parameter should rather be a cursor I guess

DaughterOfMars: I'm pretty sure the resulting SQL query doesn't care about order between the filter and the limit, because the limit always applies to the resulting set (after the filter, in other words).

Thoralf-M: Ah I thought it would already be executed earlier, but it's just done at the end in the .load(), all good then, thanks

Thoralf-M (approved):
http://localhost:3030/auctions
["000.iota","001.iota","002.iota","003.iota","004.iota","005.iota","006.iota","007.iota","008.iota","009.iota","010.iota","011.iota","012.iota","013.iota","014.iota","015.iota","016.iota","017.iota","018.iota","019.iota","020.iota","021.iota","022.iota","023.iota","024.iota","025.iota","026.iota","027.iota","028.iota","029.iota","030.iota","031.iota","032.iota","033.iota","034.iota","035.iota","036.iota","037.iota","038.iota","039.iota","040.iota","041.iota","042.iota","043.iota","044.iota","045.iota","046.iota","047.iota","048.iota","049.iota"]

http://localhost:3030/auctions?page=1
["050.iota","051.iota","052.iota","053.iota","054.iota","055.iota","056.iota","057.iota","058.iota","059.iota","060.iota","061.iota","062.iota","063.iota","064.iota","065.iota","066.iota","067.iota","068.iota","069.iota","070.iota","071.iota","072.iota","073.iota","074.iota","075.iota","076.iota","077.iota","078.iota","079.iota","080.iota","081.iota","082.iota","083.iota","084.iota","085.iota","086.iota","087.iota","088.iota","089.iota","090.iota","091.iota","092.iota","093.iota","094.iota","095.iota","096.iota","097.iota","098.iota","099.iota"]

http://localhost:3030/auctions?sort=asc&sortBy=bid&pageSize=2
["auc.iota","auct.iota"]

http://localhost:3030/auctions?search=10
["010.iota","100.iota","101.iota","102.iota","103.iota","104.iota","105.iota","106.iota","107.iota","108.iota","109.iota","110.iota"]

http://localhost:3030/auctions?search=10&sort=asc&sortBy=name
["110.iota","109.iota","108.iota","107.iota","106.iota","105.iota","104.iota","103.iota","102.iota","101.iota","100.iota","010.iota"]

Alex6323 (approved):

indexer/src/db/mod.rs
```diff
@@ -4,3 +4,45 @@
...
+    Asc,
+    Desc,
+}
+
+impl std::str::FromStr for SortOrder {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Ok(match s {
+            "asc" => SortOrder::Desc,
```
thibault-martinez: 🤔

indexer/src/api.rs
```diff
@@ -76,6 +80,74 @@ async fn get_names_for_address(
...
+    page: Option<usize>,
+    page_size: usize,
+    sort: SortOrder,
+    sort_by: AuctionSortBy,
+    search: Option<String>,
+}
+
+#[derive(Clone, Deserialize, Default)]
+#[serde(default, deny_unknown_fields, rename_all = "camelCase")]
+struct AuctionsPaginationQuery {
```
thibault-martinez: Why not only one struct?

indexer/src/db/queries.rs
```diff
@@ -92,6 +93,32 @@ pub fn get_names_for_bidder_address(
...
+pub fn get_auctions(
+    conn: &mut SqliteConnection,
+    page: Option<usize>,
+    page_size: usize,
+    sort: SortOrder,
+    sort_by: AuctionSortBy,
+    search: Option<String>,
+) -> Result<Vec<String>> {
+    let mut query = names::table
+        .inner_join(name_bids::table)
```
thibault-martinez: What are you trying to achieve with this? Bit weird

indexer/src/db/mod.rs
DaughterOfMars: How did I miss this 😆

indexer/src/api.rs
DaughterOfMars: One is for the parsing and one is the actual final usable struct. However, I can probably move this inside the parse fn.

indexer/src/db/queries.rs
DaughterOfMars: What's weird about it? It needs the bid amount for sorting.

thibault-martinez: That's the number of bids, not the bid amount

DaughterOfMars: oh :c

Thoralf-M: I thought that's wanted, because that way one can see which names seem interesting for multiple addresses, the bid amount is obviously not stored at the moment

Thoralf-M: Ok but Mario wanted  “name” or “price”

indexer/src/api.rs
```diff
@@ -76,6 +80,74 @@ async fn get_names_for_address(
...
+
+        let sort_by = query
+            .sort_by
+            .as_deref()
+            .map_or(Ok(Default::default()), str::parse)
+            .map_err(|e: anyhow::Error| ApiError::BadRequest(e.to_string()))?;
+
+        Ok(Self {
+            page: query.page,
+            page_size: query.page_size.unwrap_or(50),
```
Thoralf-M: Do we maybe want to make sure that the page size can't exceed 50?

Thoralf-M:
Second round, I created max-bidding.iota with the largest bid amount and min-bidding.iota with smallest amount and the others in between:
http://localhost:3030/auctions
["000.iota","001.iota","002.iota","003.iota","004.iota","005.iota","006.iota","007.iota","008.iota","009.iota","010.iota","max-bidding.iota","min-bidding.iota"]
http://localhost:3030/auctions?sort=asc&sortBy=bid&pageSize=2
["min-bidding.iota","000.iota"]
http://localhost:3030/auctions?sort=desc&sortBy=bid&pageSize=2
["max-bidding.iota","010.iota"]
http://localhost:3030/auctions?search=1
["001.iota","010.iota"]

indexer/src/api.rs
DaughterOfMars: Hmm. I'm not sure. I'll ask Mario?

DaughterOfMars: No preference, so I vote that we just leave it for now.

Thoralf-M: If someone spams the API with a huge pageSize it could cause problems once there are many auctions

DaughterOfMars: Alright, fair enough, I'll limit it then

Thoralf-M (approved):

indexer/src/db/queries.rs
DaughterOfMars: we good on this?

thibault-martinez (changes requested):
Testing and reviewing this

indexer/src/api.rs
```diff
@@ -76,6 +80,74 @@ async fn get_names_for_address(
...
+
+async fn get_auctions(
+    State(state): State<ApiState>,
+    AuctionsPagination {
+        page,
+        page_size,
+        sort,
+        sort_by,
+        search,
+    }: AuctionsPagination,
```
thibault-martinez, line 142: Maybe another PR but it probably makes sense to offer the same options for `get_names_for_address` (which btw I think we should rename)

DaughterOfMars: Sure, I'll do it in the next PR

indexer/src/db/queries.rs
```diff
@@ -43,29 +44,32 @@ pub fn get_or_create_name(conn: &mut SqliteConnection, name: &str) -> Result<Nam
...
+pub fn add_bids_entry(
     conn: &mut SqliteConnection,
     bidder_address: &str,
     name_str: &str,
+    bid: u64,
 ) -> Result<()> {
     let bidder = get_or_create_bidder(conn, bidder_address)?;
     let name = get_or_create_name(conn, name_str)?;
-    create_bidder_name_relationship(conn, bidder.id, name.id)
+    create_bid(conn, bidder.id, name.id, bid)
```
thibault-martinez: I'd just merge `create_bid` into `add_bids_entry` tbh, we never use it elsewhere and it's not even a nice API to work with since it takes DB IDs

indexer/src/api.rs
thibault-martinez: https://github.com/iotaledger/iota-names/issues/515

indexer/src/db/models.rs
```diff
@@ -20,20 +20,12 @@ pub struct Bidder {
 #[derive(Identifiable, Selectable, Queryable, Associations, Debug)]
 #[diesel(belongs_to(Bidder))]
 #[diesel(belongs_to(Name))]
-#[diesel(table_name = bidder_name)]
-#[diesel(primary_key(bidder_id, name_id))]
-pub struct BidderName {
+#[diesel(table_name = bids)]
+#[diesel(primary_key(name_id, bidder_id, bid))]
```
thibault-martinez, line 24: I think we can add a unique clause without making the primary key too complex, I think this degrades performances, but I'm not too sure

indexer/src/db/queries.rs
```diff
@@ -84,33 +88,55 @@ pub fn get_names_for_bidder_address(
...
+        .offset((page.unwrap_or_default() * page_size) as _)
+        .into_boxed();
+    if let Some(search) = search {
+        query = query.filter(names::name.like(format!("%{search}%")))
+    }
+    query = match (sort, sort_by) {
+        (SortOrder::Asc, AuctionSortBy::Name) => query.order(names::name.asc()),
+        (SortOrder::Desc, AuctionSortBy::Name) => query.order(names::name.desc()),
+        (SortOrder::Asc, AuctionSortBy::Bid) => {
+            query.order((dsl::max(bids::bid).asc(), names::id.asc()))
```
thibault-martinez, line 111: Can you explain the `max` please?

thibault-martinez, line 111: Why the name ID order as well?

indexer/src/db/queries.rs
```diff
@@ -84,33 +88,55 @@ pub fn get_names_for_bidder_address(
...
+        }
+        (SortOrder::Desc, AuctionSortBy::Bid) => {
+            query.order((dsl::max(bids::bid).desc(), names::id.desc()))
+        }
+    };
+    Ok(query
+        .load::<(Name, Option<i64>)>(conn)?
```
thibault-martinez, line 118: Just curious as to from which command the options comes from, group_by, limit, offset?

indexer/src/db/models.rs
```diff
@@ -20,20 +20,12 @@ pub struct Bidder {
...
+pub struct Bids {
```
thibault-martinez: The plural is inconsistent with the others here. What we've done so far is that the table is plural but the type (which is a single entry) is singular.

indexer/src/db/queries.rs
DaughterOfMars: This is the name of the column we're sorting by, the max of the bids for each bidder

DaughterOfMars: Just to ensure consistency between forward and reverse order

indexer/src/db/queries.rs
DaughterOfMars: `max` actually

indexer/src/db/models.rs
DaughterOfMars: Okay, I'll fix it

indexer/src/db/models.rs
DaughterOfMars: Tbh I'm not entirely sure what the best solution is here

thibault-martinez (approved):

PR #501, opened 2025-07-23
feat(sdk): Add validateCoupon(s) with tests

### LINKED ISSUE #492: Add coupon validation in SDK
Start from branch https://github.com/iotaledger/iota-names/pull/491

- Implement the coupons validation missing [here](https://github.com/iotaledger/iota-names/pull/491/files#diff-a9bac152ce35c823f8ae4b7447382f03ab3437980a44b2806209d5b847e9d66bR67)

- Implement a suite of tests for the coupons validation code

KeitoTadashi:
fixes #492


sdk/src/utils.ts
```diff
@@ -63,6 +63,81 @@ export function validateIotaName(
...
+): boolean {
+    if (!rules.years || rules.years === null) {
+        return true;
+    }
+    const { from: minYears, to: maxYears } = rules.years;
+    return years >= minYears && years <= maxYears;
+}
+
+export function isValidCouponPercentage(amount: string): boolean {
+    return parseFloat(amount) > 0 && parseFloat(amount) <= 100;
```
msarcev: Any reason for `parseFloat` ? Afaik these come from rust `u64` so they should be integers.

sdk/src/utils.ts
```diff
@@ -63,6 +63,81 @@ export function validateIotaName(
...
+}
+
+export function isCouponExpired(
+    rules: { expiration?: string | null },
+    currentTimestamp: number = Date.now(),
+): boolean {
+    if (!rules.expiration || rules.expiration === null) {
+        return false;
+    }
+    return currentTimestamp > Number(rules.expiration);
```
msarcev: This should work, but we need to make sure both timestamps are the same units (seconds vs milliseconds)

msarcev (commented):
Please make the following changes:

- Move these functions to a `validate.ts`
- Instead of returning booleans, make them `throw` the appropriate error for each (you can see the errors in `rules.move`)

sdk/src/utils.ts
KeitoTadashi: u right should be `parseInt`

sdk/src/validate.ts
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) Mysten Labs, Inc.
+// Modifications Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+const EInvalidYears = 'Coupon is not valid for the given number of years.';
```
msarcev: Please rename these to `INVALID_YEARS` etc..

sdk/src/validate.ts
```diff
@@ -0,0 +1,82 @@
...
+    if (parseInt(rules.available_claims) > 0) {
+        return true;
+    }
+    throw new Error(EInvalidAvailableClaims);
+}
+
+export function isCouponValidForNameYears(
+    rules: { years?: { from: number; to: number } | null },
+    years: number,
+): boolean {
```
msarcev: I see no point to return a boolean if it's always going to be `true`.
I suggest it be a `void` return type and just let it throw the error if validation fails.

msarcev (approved):

PR #502, opened 2025-07-23
fix(indexer): Add `temp-subname-proxy` package to IOTA-Names package list

DaughterOfMars:
## Description

Adds the `temp-subname-proxy` package to the list of valid IOTA-Names packages and exports it in the util envs script.

Fixes an issue where subnames were not included in the following graph:

image

### DIFF
--- a/indexer/src/config.rs
+++ b/indexer/src/config.rs
@@ -16,6 +16,8 @@ pub struct IotaNamesExtendedConfig {
     pub coupons_package_address: IotaAddress,
     /// Address of the `subnames` package.
     pub subnames_package_address: IotaAddress,
+    /// Address of the `temp-subname-proxy` package
+    pub subname_proxy_package_address: IotaAddress,
     /// ID of the `AuctionHouse` object.
     pub auction_house_id: ObjectID,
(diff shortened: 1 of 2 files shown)


indexer/src/config.rs
```diff
@@ -69,31 +74,38 @@ impl IotaNamesExtendedConfig {
...
-            || package_address == self.subnames_package_address
+        [
+            self.auction_package_address,
+            self.coupons_package_address,
+            self.iota_names_config.package_address,
+            self.iota_names_config.payments_package_address,
+            self.subnames_package_address,
+            self.subname_proxy_package_address,
+        ]
+        .contains(&package_address.into())
```
Thoralf-M: @thibault-martinez didn't like this change, because it creates the array every time the fn is called

Thoralf-M (approved):

indexer/src/config.rs
thibault-martinez: I definitely prefer the other version

DaughterOfMars: fine, I reverted it

thibault-martinez (approved):

PR #505, opened 2025-07-23
Show errors correctly

### LINKED ISSUE #503: Exceed length in availableCheck should show length error
Image

Error should be: `Name must be 3-63 characters long`

KeitoTadashi:
fixes #503


Juligs (approved):

msarcev (changes requested):
How come this issue wasn't caught in the tests of `validateIotaName` ?
We should probably cover it with some tests

msarcev (approved):

PR #508, opened 2025-07-23
feat: add coupon ui for dapp

### LINKED ISSUE #496: Add coupon support for DAPP
Inspiration from figma

Image

evavirseda:
fixes https://github.com/iotaledger/iota-names/issues/496


msarcev (dismissed):
Needs renew

dapp/src/components/dialogs/PurchaseNameDialog.tsx
```diff
@@ -174,10 +197,62 @@ export function PurchaseNameDialog({ name, open, setOpen, onPurchase }: Purchase
...
+
+            try {
+                const discount = await iotaNamesClient.calculateDiscountedPrice({
+                    coupons: couponCodes,
+                    name,
+                    years: renewYears,
+                    isRegistration: true,
+                });
+
+                console.log('Discounted price:', discount);
```
KeitoTadashi: 🧹

KeitoTadashi (approved):
Good work! 💪

sdk/src/iota-names-client.ts
```diff
@@ -491,7 +491,14 @@ export class IotaNamesClient {
             coupons = (await Promise.all(couponPromises)) as Coupon[];
         }
 
-        validateCoupons(coupons, years, name.length, address);
+        const nameParts = normalizeIotaName(name, 'dot').split('.');
+        const firstNamePart = nameParts[0];
+
+        if (nameParts.length < 2 || !firstNamePart) {
+            throw new Error('Invalid IOTA name.');
```
marc2332: Why was this added here? Looks like out of place

dapp/src/components/CouponsWrapper.tsx
marc2332, line 1: Can we name this something more meaningful? Like `CouponsInput` maybe?

marc2332 (dismissed):

sdk/src/iota-names-client.ts
VmMad: why does it look out of place? if it can't be resolved like this it's because the name is invalid, isn't it?

dapp/src/components/CouponsWrapper.tsx
VmMad: Thought the same, but bthe reaosn is that it's not only the input, but also the chips for the coupons, so maybe `CouponInputSection`?

sdk/src/iota-names-client.ts
marc2332: Yeah but, why in this method and not the rest? Why only this one? Is like having 10 children of all the same birthdate and only give a gift to the third child and not even congratulating the rest

marc2332: Do we even need these methods to be validating the name? And why is the validation custom, why not use `validateIotaName` ?

dapp/src/components/CouponsWrapper.tsx
marc2332: Section implies the existence of other sections, maybe `CouponInputSelection` ?

sdk/src/iota-names-client.ts
VmMad: there are other methods that also validate user input, this is not the only one, but yeah we can use validateIotaName

VmMad (approved):

msarcev (approved):

PR #513, opened 2025-07-25
Add names-display deploy workflow

shufps:
Adds manual trigger for docker built and pushes to a private registry


.github/workflows/docker-deployment.yaml
thibault-martinez, line 1: I think the workflow name should be a bit more explicit, we have other docker stuff in this repo

.github/workflows/docker-deployment.yaml
```diff
@@ -0,0 +1,53 @@
+name: Deploy Workflow
```
thibault-martinez: Same for this

msarcev (approved):

PR #521, opened 2025-07-28
feat(indexer): move AUCTIONS_DB in data folder

Thoralf-M:
Move AUCTIONS_DB in data folder so it gets persisted with docker setup

Tested by running `cargo run start`


shufps (approved):
lgtm

PR #534, opened 2025-07-29
feat(dapp): improve name texts in image-display

### LINKED ISSUE #531: Improve name texts in `image-display`
Image

Currently some nfts have text that overlow the svg, we should try to avoid this by playing with:
- font sizes
- line breaks
- ellipsis

- Avoid repeating the name in the svg if no subname is present. currently it shows as `name@name` if the nft is for `name.iota`
- Make all text uppercase, and use roboto flex if not used already. 
- Try to keep the subname on top of the name, not on the same line. It seems that when a subname is present, the text size of the name is smaller.

Image

Edit:
* A top level name should be rendered by wrapping up to four lines and then truncated
* If it's a sub name (i.e. `test.one.two.three.iota`). The subnames, up to the top level name are rendered by wrapping, while the top level name is rendered below with `@three`.
 Like so

Juligs:
fixes #531


msarcev (approved):
🤩

msarcev (approved):

VmMad (approved):

marc2332 (approved):
lgtm

panteleymonchuk (approved):

PR #536, opened 2025-07-30
feat(indexer): warn logs for iota_data_ingestion_core

Thoralf-M:
Log errors from iota_data_ingestion_core to help debugging in case something fails


Alex6323 (approved):

indexer/docker/docker-compose.yml
```diff
@@ -26,7 +26,7 @@ services:
     volumes:
       - indexer_progress:/data
     environment:
-      - RUST_LOG=info,iota_names_indexer=debug,iota_data_ingestion_core=off
+      - RUST_LOG=info,iota_names_indexer=debug,iota_data_ingestion_core=error
```
thibault-martinez: Depending on their judgement of what is suspicious, maybe even `warn` would be interesting idk

thibault-martinez (approved):

indexer/docker/docker-compose.yml
Thoralf-M: @sergiupopescu199 any recommendation?

sergiupopescu199: @thibault-martinez @Thoralf-M 

My advise will be to include `info` logs from the ingestion core, it helps a lot during debug and tracking the overall execution of the checkpoint pipeline flow, but if it's too verbose then `warn` is also very good!

Thoralf-M: It's very verbose, all the checkpoints are logged twice with that on info, that's why we disabled it in the first place

Thoralf-M: Changed to `warn`

sergiupopescu199: I can understand, it logs when a worker starts processing and finishes the same checkpoint, but on the other hand it helped us debug quite easily some non trivial scenarios in the indexer itself.

I believe `warn` will be more than enough in this case!

PR #546, opened 2025-07-31
fix(dapp): fetch display for auctions

### LINKED ISSUE #535: Infinite loading of display image for auctions
Currently in the `dapp` the auctions are not able to resolve the display image through the logic in `AvatarDisplay`

Solve this by calling the `names-display` directly

- Add an `ENV` variable to configure the names-display API base URL
- Add a hook to fetch an image based on name and expirationDate
- Use this hooks in the places where we render auction "NameCard"(s)

KeitoTadashi:
fixes #535


dapp/src/auctions/hooks/useAuctionDisplayImages.ts
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+import { useQuery } from '@tanstack/react-query';
+
+export function useAuctionDisplayImages(name: string, expirationTimestamp: string) {
```
VmMad:
```suggestion
export function useAuctionDisplayImage(name: string, expirationTimestamp: string) {
```

dapp/src/auctions/hooks/useAuctionDisplayImages.ts
VmMad, line 1: wouldn't it be better to just create a function to get the url and then pass it to an `` directly ?

dapp/src/components/name-card/NameCard.tsx
```diff
@@ -35,6 +37,7 @@ export function NameCard({
                 size={size}
                 badge={badge}
                 button={menuOptions ? <ContextMenuButton options={menuOptions} /> : null}
+                fallbackUrl={auctionDisplay}
```
VmMad: i think it would be better to show add another prop to receive an static image url and render an `<img` based on that, so that we avoid trying to fetch registration data for bids just to try to load the image, which we know will return undefined.

VmMad: this prop is used as a fallback if the avatar and image display are not available, so it would try to load both and at the end set this string, so i feel like it would be better to use another prop

dapp/next.config.mjs
```diff
@@ -4,6 +4,7 @@
 import { execSync } from 'child_process';
 
 let NEXT_PUBLIC_IOTA_NAMES_REV = 'development';
+let NEXT_PUBLIC_NAMES_DISPLAY_API_URL = process.env.NEXT_PUBLIC_NAMES_DISPLAY_API_URL;
```
msarcev:
```suggestion
let NEXT_PUBLIC_NAMES_DISPLAY_API_URL = '';
```

dapp/src/app/(protected)/my-names/components/PanelTile.tsx
```diff
@@ -57,7 +57,7 @@ export function PanelTile({
                                 type={ImageType.BgTransparent}
                                 shape={ImageShape.SquareRounded}
                             >
-                                <AvatarDisplay
+                                <NameCardDisplay
```
msarcev: I don't think this needs to change.

dapp/src/auctions/hooks/useAuctionDisplayImage.ts
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+import { useQuery } from '@tanstack/react-query';
+
+export function useAuctionDisplayImage(name: string, expirationTimestamp: string) {
+    return useQuery({
```
msarcev: We probably don't need react query to build this string. A util function should be fine.

dapp/next.config.mjs
```diff
@@ -15,10 +16,17 @@ try {
     console.warn('Could not get git revision, using default');
 }
 
+if (process.env.NEXT_PUBLIC_NAMES_DISPLAY_API_URL) {
+    NEXT_PUBLIC_NAMES_DISPLAY_API_URL = process.env.NEXT_PUBLIC_NAMES_DISPLAY_API_URL;
+} else {
+    throw new Error('NEXT_PUBLIC_NAMES_DISPLAY_API_URL env must be configured');
+}
+
```
VmMad: you've already created the `let` with the `.env`
```suggestion
if (!NEXT_PUBLIC_NAMES_DISPLAY_API_URL) {
    throw new Error('NEXT_PUBLIC_NAMES_DISPLAY_API_URL env must be configured');
}

```

dapp/src/components/name-record/NameCardDisplay.tsx
msarcev, line 1: I don't think we need all these changes. We should probably leave `AvatarDisplay` unchanged.

dapp/src/components/name-card/NameCard.tsx
```diff
@@ -30,11 +32,12 @@ export function NameCard({
                 nftDisplayVariants({ size }),
             )}
         >
-            <AvatarDisplay
+            <NameCardDisplay
```
msarcev: Instead of doing this `NamesDisplay` changes. Just handle it here
```
auctionDisplay ? <img .... > : <AvatarDisplay .... />
```

msarcev (changes requested):

dapp/src/auctions/hooks/useAuctionDisplayImage.ts
```diff
@@ -0,0 +1,22 @@
...
+            }
+            const url = `${baseUrl}/${name}/${expirationTimestamp}`;
+            return {
+                url,
+            };
+        },
+        enabled: !!name && !!expirationTimestamp,
+        gcTime: 0,
+    });
+}
```
VmMad: There's no real benefit of doing this as a query, its better to do a function instead. 
Also the env should already be defined at this point

```ts
function getNameDisplaySrc(name, expirationTimestamp) {
    const baseUrl = process.env.NEXT_PUBLIC_NAMES_DISPLAY_API_URL;
    return new URL(`${name}/${expirationTimestamp}`, baseUrl).href;
}
```

dapp/src/components/name-card/NameCard.tsx
```diff
@@ -30,11 +32,12 @@ export function NameCard({
...
                 name={name}
                 size={size}
                 badge={badge}
                 button={menuOptions ? <ContextMenuButton options={menuOptions} /> : null}
+                fallbackUrl={auctionDisplay}
```
VmMad: if we dont use `staticImageSrc` then the component will go through the avatar route and execute unnecessary hooks
```suggestion
                staticImageSrc={auctionDisplay}
```

dapp/src/components/name-card/NameCard.tsx
```diff
@@ -7,11 +7,12 @@ import { nftDisplayVariants } from '@/components/name-record/variants';
 import { MenuListItem, type NftDisplayProps } from '@/lib/types/components';
 
 import { ContextMenuButton } from '../buttons/ContextMenuButton';
-import { AvatarDisplay } from '../name-record/AvatarDisplay';
+import { NameCardDisplay } from '../name-record/NameCardDisplay';
 
 interface NameCardProps extends NftDisplayProps {
     menuOptions?: MenuListItem[];
     isSelected?: boolean;
+    auctionDisplay?: string;
```
VmMad: this prop can be agnostic to `auction`, so i'd change it to something like `displaySrc`

VmMad (dismissed):
Lint is failing 🙏🏽

dapp/src/components/name-record/NameCardDisplay.tsx
VmMad: I suggested this change. This was so that if you passed a `staticImageSrc` it didn't need to load or fetch any unnecessary things e.g (useRegistrationNfts inside the avatarDisplay). But you're correct it can be handled from outside

dapp/src/components/name-card/NameCard.tsx
VmMad: or yes... this could be a solution too. handling from here and leaving the AvatarDisplay as it was

dapp/src/components/name-record/AvatarDisplay.tsx
```diff
@@ -15,38 +14,40 @@ import { nftDisplayVariants } from './variants';
...
 }
 
-export function AvatarDisplay({ name, size, badge, button, fallbackUrl }: AvatarDisplayProps) {
+export function AvatarDisplay({
+    name,
+    size,
+    badge,
+    button,
+    fallbackUrl,
+    staticImageSrc,
```
msarcev: I don't see why this is needed ? 
@VmMad what was your idea ?

msarcev (approved):

marc2332 (approved):

PR #565, opened 2025-08-06
feat(iota-names): remove admin registered names from the reserved deny list

### LINKED ISSUE #524: Remove names from reserved list if registered by admin
If we don't remove it, usage would be limited as users can't register sub names
So calling `admin::{register_name, register_names}` should also remove a potential entry from the reserved list

Thoralf-M:
Remove admin registered names from the reserved deny list, to make them useable. If they would stay in the deny list, they couldn't be used to create subdomains for example.

Fixes https://github.com/iotaledger/iota-names/issues/524

Added new tests


packages/iota-names/sources/deny_list.move
```diff
@@ -97,6 +97,39 @@ public fun remove_blocked_names(iota_names: &mut IotaNames, _: &AdminCap, labels
     );
 }
 
+/// Removes all reserved labels for a given Name (except top-level)
```
DaughterOfMars:
```suggestion
/// Removes all reserved labels for a given name (except top-level)
```

packages/iota-names/sources/deny_list.move
```diff
@@ -97,6 +97,39 @@ public fun remove_blocked_names(iota_names: &mut IotaNames, _: &AdminCap, labels
...
+        let label = name.label(index);
+        if (reserved_table_mut.contains(*label)) {
+            reserved_table_mut.remove(*label);
+        };
+        index = index + 1;
+    };
+}
+
+/// Removes all reserved labels for a vector of names (except top-level)
+public(package) fun remove_reserved_labels_for_names(iota_names: &mut IotaNames, names: &vector<Name>) {
```
DaughterOfMars, line 115: any reason not to call the other fn?

Thoralf-M: Wanted to call `reserved_table_mut()` only once, with the other fn it would be called for each name

DaughterOfMars: alright, fair enough

DaughterOfMars (approved):

packages/iota-names/sources/admin.move
```diff
@@ -40,17 +43,25 @@ public fun register_name(
...
     assert!(!names.is_empty(), ENoNamesProvided);
     let sender = sender(ctx);
     let config = *iota_names.get_config<CoreConfig>();
+    let mut name_structs = vector::empty<Name>();
+    let mut tmp_names = names;
+    while (!tmp_names.is_empty()) {
+        let n = name::new(tmp_names.pop_back());
+        name_structs.push_back(n);
+    };
+    deny_list::remove_reserved_labels_for_names(iota_names, &name_structs);
```
Alex6323, line 59: I don't quite understand it yet. Why do we need to map and reverse the order here? Seems like we could just pass a reference to the `vector<String>` to `remove_reserved_labels_for_names`?

Thoralf-M: The name we need to construct anyways for the .add_record() and in `remove_reserved_labels_for_names()` we need to get access to the labels, so we would need to add custom parsing if we don't go through the name
order doesn't matter

Alex6323: okay then ... but the code feels like it could be avoided, but not so sure.

Alex6323 (approved):

packages/iota-names/sources/admin.move
Thoralf-M: What could be avoided?

packages/iota-names/sources/admin.move
```diff
@@ -10,10 +10,12 @@ use iota::clock::Clock;
 use iota::tx_context::sender;
 use iota_names::core_config::CoreConfig;
 use iota_names::name;
+use iota_names::name::Name;
```
Thoralf-M:
```suggestion
use iota_names::name::{Self, Name};
```

PR #566, opened 2025-08-06
feat(names-display): add expired bg to names-display

### LINKED ISSUE #549: Add expired background to names-display
When the timestamp received is already in the past, we should select another base svg for the background. 

Image

Juligs:
fixes #549


marc2332 (approved):
🟥 ✔️

VmMad (approved):

PR #579, opened 2025-08-08
feat(indexer): Add `auctions` table to track auction statuses

DaughterOfMars:
## Description

Adds a new table to track the statuses of auctions so that the API endpoints can query based on them.

Auctions can be in one of three states:
- Active
- Finished
- Claimed

Which can be queried in both auction endpoints via the `status` variable.


Thoralf-M (approved):

marc2332 (approved):
looking good

PR #600, opened 2025-08-12
fix(indexer): Use milliseconds for now timestamp

DaughterOfMars:
## Description

The expiration timestamp is in millis, but currently the API uses the UNIX timestamp to check whether the auctions are finished.


marc2332 (approved):
lgtm

msarcev (approved):

thibault-martinez (approved):

PR #610, opened 2025-08-13
fix(dapp): Fix redirect when landing in  `/auctions`

### LINKED ISSUE #590: Intermittent redirect when opening protected pages directly or reload
Sometimes, when opening a protected page (like /my-names) directly in the browser, the app incorrectly redirects to the wallet connect page even if the wallet is already connected.

Key points:
- Happens on direct navigation or on reload.
- Issue occurs intermittently. Issues is not reproducing with slow internet mode.
- Likely related to how the page checks connection status before it’s fully loaded.
- Needs adjustment so that the check waits until connection status is confirmed.
- Please, check `ProtectedLayout`

https://github.com/user-attachments/assets/d24ec5ad-ec94-41e9-a8ba-86aa3aeeb1af

KeitoTadashi:
fixes #590

### DIFF
--- a/dapp/src/app/(protected)/layout.tsx
+++ b/dapp/src/app/(protected)/layout.tsx
@@ -3,7 +3,7 @@
 
 'use client';
 
-import { useCurrentAccount } from '@iota/dapp-kit';
+import { useCurrentAccount, useCurrentWallet } from '@iota/dapp-kit';
 import { redirect } from 'next/navigation';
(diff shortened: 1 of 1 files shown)


panteleymonchuk (approved):
checked, works properly 🚀

marc2332 (approved):

msarcev (approved):

PR #631, opened 2025-08-14
feat(dapp): Fix Renew Dialog expiration date

### LINKED ISSUE #630: Compute expiration date when renewing a name or subname
Image

KeitoTadashi:
fixes #630


dapp/src/components/dialogs/RenewNameDialog.tsx
```diff
@@ -321,13 +321,40 @@ export function RenewNameDialog({ setOpen, name, onRenew }: RenewDialogProps) {
...
+                                        icon={isLoading ? <LoadingIndicator /> : null}
+                                        label="Registration Expires"
+                                        value={(() => {
+                                            const parentObject = getParentObject(
+                                                ownedNames,
+                                                ownedSubnames,
+                                                nameRecord.nameRecord.name,
+                                            );
+                                            return formatExpirationDate(
+                                                new Date(parentObject?.expirationTimestampMs || 0),
```
marc2332: What if the parent object can be null? mmm 

can it?

KeitoTadashi: a subname should always have a parent, i don't think it can be null

VmMad: it probably can, if you get sent a subname you don't own the parent name so you wont find it

KeitoTadashi: can we renew a subname with a non owned parent?


```ts
    // Renew subnames
    if (isNameSubname && nameRecord && namePermissions?.allowTimeExtension && !isExpired) {
        const objectId = getNameObject(ownedSubnames, nameRecord.name);
        const parentObject = getParentObject(ownedNames, ownedSubnames, nameRecord.name);
        if (objectId && parentObject) {
            // Only allow extending the expiration time if its less than its parent
            const expiresBeforeParent =
                nameRecord.expirationTimestampMs < parentObject?.expirationTimestampMs;
            if (expiresBeforeParent) {
                updates.push({
                    type: 'renew-subname',
                    nftId: objectId,
                    expirationTimestampMs: parentObject.expirationTimestampMs,
                });
            }
        }
    }
    ``

**Review by marc2332 — CHANGES_REQUESTED** [review]
Can you move your logic from outside the jsx to L265? That's where it makes more sense imo

**Review by marc2332 — CHANGES_REQUESTED** [review]
Image

its not being formatted, let me fix this

**Review by marc2332 — APPROVED** [review]

File: dapp/src/components/dialogs/RenewNameDialog.tsx [hunk 2]
```diff
@@ -262,9 +262,27 @@ export function RenewNameDialog({ setOpen, name, onRenew }: RenewDialogProps) {
 
     const wantsToRenew = isNameSubname || !!renewYears;
     const canRenew = nameRecord && updates.length > 0;
-    const expirationDate = nameRecord?.nameRecord.expirationTimestampMs
-        ? formatExpirationDate(new Date(nameRecord.nameRecord.expirationTimestampMs))
-        : null;
+    const expirationDate = (() => {
+        if (nameRecord?.nameRecord) {
+            if (isNameSubname && ownedNames && ownedSubnames) {
+                // Subnames
```
VmMad: imo these comments are not necessary

VmMad (approved):

PR #633, opened 2025-08-18
feat: show carousel on landing

### LINKED ISSUE #614: Implement Live Auctions Landing carousel
Check with designs and implement the Live Auctions component visible on Landing in figma 

Image

panteleymonchuk:
Closes: #614


dapp/src/sections/AuctionCarousel.tsx
```diff
@@ -0,0 +1,160 @@
...
+        breakpoint: { max: 1024, min: 768 },
+        items: 2,
+        slidesToSlide: 1,
+    },
+    mobile: {
+        breakpoint: { max: 768, min: 0 },
+        items: 1,
+        slidesToSlide: 1,
+    },
+};
```
Juligs: - Across all versions, the image size could be limited with max-h and max-w of 220px like in Figma  to prevent the card from growing too much :pray: 

image

- On mobile, maybe the card could be centered, because the content doesn’t look very clear.

- Tablet version 

image

dapp/src/sections/AuctionCarousel.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+            <div className="relative">
+                <div className="absolute top-0 left-0 w-[60px] md:w-[120px] lg:w-[150px] h-full bg-gradient-to-r from-[#0b0c23] via-[#0b0c23cc] to-transparent pointer-events-none z-10"></div>
+                <div className="absolute top-0 right-0 w-[60px] md:w-[120px] lg:w-[150px] h-full bg-gradient-to-l from-[#0b0c23] via-[#0b0c23cc] to-transparent pointer-events-none z-10"></div>
+
+                <Carousel
+                    responsive={responsive}
+                    infinite={true}
+                    autoPlay={true}
+                    autoPlaySpeed={3000}
```
Juligs: [Screencast from 2025-08-20 12-51-35.webm](https://github.com/user-attachments/assets/80d15ca0-5a6c-4434-a9ca-93bde1ebc0e8)

- Does autoplay start automatically, or only after interacting with the slider? currently it only starts on interaction 

-  Near the end of the loop, the first card glitches and jumps back to its initial position :thinking:

Juligs (changes requested):
LGTM :star2: , just a few things :

dapp/src/sections/AuctionCarousel.tsx
panteleymonchuk: > Does autoplay start automatically, or only after interacting with the slider? currently it only starts on interaction

it should start automatically but pause on hover

---

> Near the end of the loop, the first card glitches and jumps back to its initial position

I also saw it. For me it was random stuff, doesn't related to the end of the list. I removed custom transition, now it should be better.

dapp/src/sections/AuctionCarousel.tsx
```diff
@@ -0,0 +1,160 @@
...
+        );
+    }
+
+    return (
+        <div className="w-full">
+            <AuctionCarouselHeader />
+
+            <style jsx global>{`
+                .react-multi-carousel-track {
+                    gap: 0;
```
Juligs:
```suggestion
                    gap: 24px;
                    padding: 8px;
```
We could add a gap and some padding so the box-shadow is more visible and there’s a separation similar to the one in Figma, like this: 
image

dapp/src/sections/AuctionCarousel.tsx
```diff
@@ -0,0 +1,160 @@
...
+                    padding: 0;
+                    display: flex;
+                    justify-content: center;
+                    min-width: 220px !important;
+                }
+                .react-multi-carousel-dot-list,
+                .react-multiple-carousel__arrow {
+                    display: none !important;
+                }
+            `}</style>
```
Juligs: Maybe we can move the CSS styles to a file, for example `globals.css`

dapp/src/sections/AuctionCarousel.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+            <div className="relative">
+                <div className="absolute top-0 left-0 w-[60px] h-full bg-gradient-to-r from-[#0b0c23] via-[#0b0c23cc] to-transparent pointer-events-none z-10"></div>
+                <div className="absolute top-0 right-0 w-[60px] h-full bg-gradient-to-l from-[#0b0c23] via-[#0b0c23cc] to-transparent pointer-events-none z-10"></div>
+
+                <Carousel
+                    responsive={responsive}
+                    infinite
+                    autoPlay
```
Juligs: [Screencast from 2025-08-21 16-45-26.webm](https://github.com/user-attachments/assets/d5ddf528-a3ea-4847-ae0f-a6dacfdb2a32)

The transition when the loop finishes is better, but you can still see a flicker at the end of the loop

Juligs (changes requested):
Looks great, we’re almost there,  just a few small things left to adjust :cat:

dapp/src/sections/AuctionCarousel.tsx
panteleymonchuk: I wouldn't do it. This auction functionality something temporary, so better to keep it here imo.
The same reason we have `src/auctions` page as some kind of separate module

dapp/src/sections/AuctionCarousel.tsx
panteleymonchuk: I see, I'll try to manage it, but it might be issue of the library

dapp/src/sections/AuctionCarousel.tsx
```diff
@@ -0,0 +1,144 @@
...
+        items: 4,
+        slidesToSlide: 1,
+    },
+    tablet: {
+        breakpoint: { max: 1024, min: 768 },
+        items: 3,
+    },
+    mobile: {
+        breakpoint: { max: 768, min: 0 },
+        items: 1,
```
brancoder: should we change this to 2 items because with only 1 it doesnt look so good?
image

brancoder (changes requested):
the text on the cards seams to change sometime, ex. when mouse hovers or if you wait and after few slides the last right card changes text

https://github.com/user-attachments/assets/1850b0a2-ca9c-48eb-a6a7-eb7233d49379

brancoder (approved):

dapp/src/sections/AuctionCarousel.tsx
```diff
@@ -0,0 +1,178 @@
...
+                                waitForTransition: true,
+                                pauseOnMouseEnter: true,
+                            }}
+                            centeredSlides
+                            loop
+                            className="mySwiper"
+                            modules={[Autoplay]}
+                        >
+                            {auctionsToRender.map((auction) => (
+                                <SwiperSlide key={auction.key}>
```
Juligs:
```suggestion
                                <SwiperSlide key={auction.key} className="pb-xs">
```
Could we add some bottom padding so the cards’ shadows are visible?

Juligs (changes requested):
LGTM , just one nit:

dapp/src/sections/AuctionCarousel.tsx
panteleymonchuk: added. Nice, really didn't notice that 👀

Juligs (approved):

msarcev (approved):

PR #658, opened 2025-08-21
feat(dapp): open renew popup when clicking renew on subname panel

### LINKED ISSUE #657: Add logic to `Renew` button in subnames panel
Image
This button will only show for names that are close to expiration (less that 1 week) or already expired.

VmMad:
Also fixes the button in the subnames panel. Now it shows as disabled if the selected name/subname doesn't allow for children creation

Fixes #657


brancoder (changes requested):
we should add a logic to not show the renew button if flag `allowTimeExtension` is `false`
image

brancoder (approved):
LGTM

dapp/src/app/(protected)/my-names/components/NamePanelTile.tsx
```diff
@@ -72,6 +81,7 @@ export function NamePanelTile({
                 footer={
                     isCloseToExpiration || isExpired ? (
                         <PanelFooter
+                            isRenewDisabled={!allowTimeExtension}
```
KeitoTadashi: should be also disabled if is grace period expired

```ts
const isNameGracePeriodExpired = isGracePeriodExpired(nameRecord);

```

KeitoTadashi (changes requested):
LGTM just a nit

KeitoTadashi (approved):

PR #664, opened 2025-08-22
feat(scripts): add remove-move-envs script

Thoralf-M:
Remove script to remove env entries to make it easier to publish to networks where an entry exists already
Tested by running
```
cd scripts
iota client switch --env devnet
pnpm ts-node remove-move-envs.ts
pnpm ts-node init/init.ts devnet
```

```
Cleaned: /home/t/Desktop/iota-names/packages/auction/Move.lock
Cleaned: /home/t/Desktop/iota-names/packages/coupons/Move.lock
Cleaned: /home/t/Desktop/iota-names/packages/iota-names/Move.lock
Cleaned: /home/t/Desktop/iota-names/packages/payments/Move.lock
Cleaned: /home/t/Desktop/iota-names/packages/subnames/Move.lock
Cleaned: /home/t/Desktop/iota-names/packages/temp-subname-proxy/Move.lock
Done.
Publishing IotaNames...
...
```

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -16,6 +16,12 @@ This will automatically publish all the packages in the correct order, collect a
 file, as well as do a full on-chain setup (creation of the registry, addition of pricelist, authorization of all apps as well as
 calling their respective setup functions).
 
+If you want to publish (not upgrade) to a network, where an [env] entry already exists in the Move.lock files of the packages, you can run the following script to remove them, otherwise publishing will fail:
+
+```bash
+pnpm ts-node remove-move-envs.ts
+```
+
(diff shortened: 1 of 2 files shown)


DaughterOfMars (commented):
Could we do this when you call publish? Maybe by checking for the published package first?

Thoralf-M:
> Could we do this when you call publish? Maybe by checking for the published package first?

Could be done, but I thought it's better to keep it separated because if there is already a package you might actually want to do an upgrade instead

DaughterOfMars (approved):

marc2332 (approved):
works

Issue #541, opened 2025-07-31
Remove `useBalanceValidation`


marc2332:
Hides complexity unnecessarily making it harder to reason and it is also potentially not great performance wise

PR #683, opened 2025-08-26
feat(dapp):  Improve handling of errors related to gas in dialogs 

### LINKED ISSUE #675: Add a "No Funds" error message when the user has insufficient funds to edit metadata
## Bug Description

Currently user is able to input metadata and click apply even when user has insufficient funds. After clicking apply the success toast is shown eventhough no metadata was changed.

Image

Add no funds info error message to inform user similar to what we have in Register name dialog⬇️

Image

### LINKED ISSUE #677: Disable Create subname button in dialog when user has no funds
## Bug Description

when user has no funds the dialog does show error message but the Create button is enabled and clicking on it show a success message. We should disable the button when no funds are available

Image

KeitoTadashi:
fixes #675
fixes #677


dapp/src/components/dialogs/EditMetadata.tsx
```diff
@@ -162,8 +171,12 @@ export function EditMetadataDialog({ name, setOpen }: EditMetadataDialogProps) {
         }));
     }
 
-    const isLoading = isUpdating || isSigning || isSaving;
-    const disableApply = isLoading || updates.length === 0;
+    const hasEnoughGas =
+        !updateNameError?.message.includes(NOT_ENOUGH_BALANCE_ID) &&
+        !updateNameError?.message.includes(GAS_BALANCE_TOO_LOW_ID);
```
brancoder: in the auctions bid dialog we have this check        `!balanceValidationError?.message.includes(INSUFFICIENT_COIN_BALANCE_ID);`
do we need to add this check here?
also could we extract this has enough balance check to some helper and reuse it in all dialogs?

brancoder (changes requested):
LG and works  just a comment about the has enough balance check

dapp/src/components/dialogs/EditMetadata.tsx
KeitoTadashi: i agree with move code to a new helper

dapp/src/lib/utils/enoughGas.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+import {
+    GAS_BALANCE_TOO_LOW_ID,
+    INSUFFICIENT_COIN_BALANCE_ID,
+    NOT_ENOUGH_BALANCE_ID,
+} from '../constants';
+
+export function enoughGas(transactionError: Error | null): boolean {
```
brancoder: the name is a bit misleading because it has nothing to do with gas but rather the gas error message. maybe we can change the implementation and naming a bit to make it easier to use
```
export function isGasBalanceError(transactionError: Error | null): boolean {
    return !!(
        transactionError?.message.includes(NOT_ENOUGH_BALANCE_ID) ||
        transactionError?.message.includes(GAS_BALANCE_TOO_LOW_ID) ||
        transactionError?.message.includes(INSUFFICIENT_COIN_BALANCE_ID)
    );
}
```
and use it  ex.

`const hasInsufficientBalance = isGasBalanceError(balanceValidationError);
`

brancoder (changes requested):

brancoder (changes requested):
conflicts 🙏

dapp/src/auctions/components/dialogs/AuctionBidDialog.tsx
```diff
@@ -174,10 +164,6 @@ export function AuctionBidDialog({ name, closeDialog, onCompleted }: AuctionBidD
             return `The value exceeds the maximum decimals (${IOTA_DECIMALS}).`;
         } else if (isBidBelowMinimum) {
             return `Bid must be ≥ ${minBidLabel}`;
-        } else if (!hasEnoughGas) {
-            return getUserFriendlyErrorMessage(NOT_ENOUGH_BALANCE_ID);
-        } else if (auctionError) {
-            return getUserFriendlyErrorMessage(auctionError);
         } else if (balanceValidationError) {
```
brancoder, line 167: 1. looks to me like we can remove the hook `useBalanceValidation` because all ti does is a dry run and than comparing of balances.
The auctionError is already thrown when there is no balance and balanceValidationError is not even thrown when user has no funds.
image

2. `useBalanceValidation` only does the dry run so we can add it to the `useAuctionBid` hook if needed like we do in ex.` https://github.com/iotaledger/iota-names/blob/develop/dapp/src/hooks/useUpdateNameTransaction.ts`

brancoder (changes requested):

dapp/src/auctions/components/dialogs/AuctionBidDialog.tsx
marc2332: Will be done in https://github.com/iotaledger/iota-names/issues/541

brancoder (approved):

marc2332 (approved):

PR #700, opened 2025-08-28
feat: redirect on protected route

### LINKED ISSUE #69: Implement proper disconnected redirect on protected routes
It was removed in https://github.com/iotaledger/iota-names/pull/659 because it was failing 50% of times

Go to `/` if the user lands in `/my-names` while being disconected

panteleymonchuk:
Closes iotaledger/ts-packages#69


begonaalvarezd:
Stale PR, we can come back to it if needd

PR #713, opened 2025-09-02
feat(scripts): transfer names to owners in init

thibault-martinez:


scripts/reserved-names/register-names.ts
```diff
@@ -43,17 +44,21 @@ export const parseCsvFile = (filePath: string): Record<string, string | undefine
...
 export const registerNames = (
     txb: Transaction,
-    names: string[],
+    batch: Record<string, string>,
     iotaNamesPackageId: string,
     adminCap: string,
     iotaNamesObjectId: string,
 ) => {
-    return txb.moveCall({
+    var names = Object.keys(batch);
```
Alex6323, line 52: I think I would merge this and the next line. The `names` binding isn't used anymore I think.

Alex6323 (approved):

scripts/reserved-names/register-names.ts
DaughterOfMars: It's used below

scripts/reserved-names/register-names.ts
```diff
@@ -73,34 +78,37 @@ export async function processNamesFileBatched(
     },
     network: string,
     client: any,
+    adminAddress: string,
 ) {
     const nameAddressPairs = parseCsvFile(filePath);
-    const allNames = Object.keys(nameAddressPairs);
-    let batch: string[] = [];
+    console.log(nameAddressPairs);
```
DaughterOfMars: Shall we remove this?

thibault-martinez: Yeah I just pushed what we did in the call, didn't clean anything yet

thibault-martinez: https://github.com/iotaledger/iota-names/pull/713/commits/8f534999f92650bd24e36f511fc3c6b4711e3264

scripts/reserved-names/register-names.ts
```diff
@@ -73,34 +78,36 @@ export async function processNamesFileBatched(
...
+        if (batchLen + nameLen + recipientLen > PURE_ARG_SIZE_LIMIT) {
             await sendNamesBatchTransaction(batch, packageInfo, network, client);
-            batch = [];
+            batch = {};
             batchLen = 0;
         }
-        batch.push(name);
+        batch[name] = recipient;
         batchLen += nameLen;
     }
```
DaughterOfMars:
```suggestion
    let seenAddresses = new Set<string>();
    for (const [name, address] of Object.entries(nameAddressPairs)) {
        var recipient = address || adminAddress;
        const nameLen = bcs.String.serializedSize(name)!;
        const recipientLen = seenAddresses.has(recipient)
            ? 0
            : bcs.Address.serializedSize(recipient)!;
        seenAddresses.add(recipient);
        if (batchLen + nameLen + recipientLen > PURE_ARG_SIZE_LIMIT) {
            await sendNamesBatchTransaction(batch, packageInfo, network, client);
            batch = {};
            batchLen = 0;
        }
        batch[name] = recipient;
        batchLen += nameLen + recipientLen;
    }
```

scripts/reserved-names/register-names.ts
```diff
@@ -73,34 +78,36 @@ export async function processNamesFileBatched(
...
+        const recipientLen = recipient.length + (Object.keys(batch).length > 0 ? 1 : 0);
+        if (batchLen + nameLen + recipientLen > PURE_ARG_SIZE_LIMIT) {
             await sendNamesBatchTransaction(batch, packageInfo, network, client);
-            batch = [];
+            batch = {};
             batchLen = 0;
         }
-        batch.push(name);
+        batch[name] = recipient;
         batchLen += nameLen;
```
Thoralf-M: this is missing the recipientLen

packages/iota-names/tests/unit/admin_tests.move
```diff
@@ -121,6 +121,7 @@ fun register_names_from_deny_list() {
         utf8(b"one.iota"),
         utf8(b"two.iota"),
     ];
+    let recipients = vector[ctx.sender(),ctx.sender()];
```
DaughterOfMars: format?

scripts/reserved-names/register-names.ts
DaughterOfMars: FYI I addressed this in my suggestion

Thoralf-M: Or actually just using recipientLen should be fine, because the names and the addresses are separated args and addresses we can expect to be longer

Thoralf-M: To handle all cases both should be tracked separated and then it should just end if either of them would exceed the size first

packages/iota-names/tests/unit/admin_tests.move
thibault-martinez: Ci is happy about the format

thibault-martinez: I don't know why tho

scripts/reserved-names/register-names.ts
DaughterOfMars: Why separate?

Thoralf-M: Because they are separated arguments to the transaction

DaughterOfMars: So is the limit per argument?

Thoralf-M: Yes

packages/iota-names/tests/unit/admin_tests.move
thibault-martinez: I manually formatted https://github.com/iotaledger/iota-names/pull/713/commits/9fb2f0d5b2ee1353ea6c6e76a15c3590d957b81b

DaughterOfMars (approved):

Thoralf-M (approved):

PR #769, opened 2025-09-11
feat(payment): add init_registration_with_years()

### LINKED ISSUE #762: payment: support registration for multiple years
It should be possible to register a name directly for multiple years, instead of just always 1 year and then extra renewal, because if the payment intent a coupon for multiple years can't be used to buy and renew for more years.
Similar to https://github.com/iotaledger/iota-names/blob/106b43eeb7747122e56b24c29b08424332a991bb/packages/iota-names/sources/sales/payment.move#L136 a `init_registration_with_years()` could be added which takes an additional years parameter and then computes the price based on the payment config for one year and renewal config for the additional years
`init_registration()` shouldn't be modified, as this wouldn't allow a package upgrade.

Thoralf-M:
Add a new init_registration_with_years() where one can register directly for multiple years. This was requested so coupons for multiple years can be used for the initial year and renewals, before this was not possible, as a coupon can only applied to a single PaymentIntent and renewal created a second one.

Fixes https://github.com/iotaledger/iota-names/issues/762

Tested with the added tests

### DIFF
--- a/packages/coupons/tests/coupons_tests.move
+++ b/packages/coupons/tests/coupons_tests.move
@@ -564,7 +564,7 @@ fun test_coupon_register(
     name: String,
     coupon_code: String,
     user: address,
-    amount: Option<u64>, // optional param to test for expected amount
+    mut amount: Option<u64>, // optional param to test for expected amount
 ) {
     scenario.next_tx(user);
     {
@@ -582,7 +582,7 @@ fun test_coupon_register(
             scenario.ctx(),
         );
         if (amount.is_some()) {
-            assert!(intent.request_data().base_amount() == amount.get_with_default(0));
+            assert!(intent.request_data().base_amount() == amount.extract());
         };
 
         return_shared(iota_names);
@@ -596,7 +596,7 @@ fun test_multi_coupon_register(
     name: String,
     mut coupon_codes: vector<String>,
     user: address,
-    amount: Option<u64>, // optional param to test for expected amount
+    mut amount: Option<u64>, // optional param to test for expected amount
 ) {
     scenario.next_tx(user);
     {
@@ -617,7 +617,7 @@ fun test_multi_coupon_register(
             );
         };
         if (amount.is_some()) {
-            assert!(intent.request_data().base_amount() == amount.get_with_default(0));
+            assert!(intent.request_data().base_amount() == amount.extract());
         };
 
         return_shared(iota_names);
@@ -632,7 +632,7 @@ fun test_coupon_renewal(
     renewal_years: u8,
     coupon_code: String,
     user: address,
-    amount: Option<u64>, // optional param to test for expected amount
+    mut amount: Option<u64>, // optional param to test for expected amount
 ) {
     scenario.next_tx(user);
     {
@@ -658,7 +658,7 @@ fun test_coupon_renewal(
             scenario.ctx(),
         );
         if (amount.is_some()) {
-            assert!(intent.request_data().base_amount() == amount.get_with_default(0));
+            assert!(intent.request_data().base_amount() == amount.extract());
         };
 
         return_shared(iota_names);
@@ -797,7 +797,7 @@ fun test_coupon_multi_year_register(
     years: u8,
     coupon_code: String,
     user: address,
-    amount: Option<u64>, // optional param to test for expected amount
+    mut amount: Option<u64>, // optional param to test for expected amount
 ) {
     scenario.next_tx(user);
     {
@@ -816,7 +816,7 @@ fun test_coupon_multi_year_register(
             scenario.ctx(),
         );
         if (amount.is_some()) {
-            assert!(intent.request_data().base_amount() == amount.get_with_default(0));
+            assert!(intent.request_data().base_amount() == amount.extract());
         };
 
         return_shared(iota_names);
@@ -831,7 +831,7 @@ fun test_multi_coupon_multi_year_register(
     years: u8,
     mut coupon_codes: vector<String>,
     user: address,
-    amount: Option<u64>, // optional param to test for expected amount
+    mut amount: Option<u64>, // optional param to test for expected amount
 ) {
     scenario.next_tx(user);
     {
@@ -853,7 +853,7 @@ fun test_multi_coupon_multi_year_register(
             );
         };
         if (amount.is_some()) {
-            assert!(intent.request_data().base_amount() == amount.get_with_default(0));
+            assert!(intent.request_data().base_amount() == amount.extract());
         };
 
         return_shared(iota_names);
--- a/packages/coupons/tests/coupons_tests.move
+++ b/packages/coupons/tests/coupons_tests.move
@@ -685,6 +685,190 @@ fun init_renewal(
 }
 
 #[test]
+fun test_coupon_multi_year_registration() {
+    let mut scenario_val = test_init();
+    let scenario = &mut scenario_val;
+    populate_coupons(scenario);
+
+    // Test 25% discount on 2-year registration (within max 2 years limit)
+    // Base price for 4-char name: 200 (registration) + 200*1 (renewal) = 400
+    // 25% discount: 400 * 0.75 = 300
+    test_coupon_multi_year_register(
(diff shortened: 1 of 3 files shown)


packages/iota-names/sources/sales/payment.move
```diff
@@ -148,6 +148,37 @@ public fun init_registration(iota_names: &mut IotaNames, name: String): PaymentI
     })
 }
 
+/// Creates a `PaymentIntent` for registering a new name for multiple years.
+/// This is a hot-potato and can only be consumed in a single transaction.
+public fun init_registration_with_years(iota_names: &mut IotaNames, name: String, years: u8): PaymentIntent {
+    let name = name::new(name);
+    validation::assert_is_valid_for_sale(iota_names.get_config<CoreConfig>(), iota_names, &name);
+    assert!(years <= iota_names.get_config<CoreConfig>().max_years(), ECannotExceedMaxYears);
```
DaughterOfMars, line 163: It can also be zero and would fail below

DaughterOfMars (commented):
What about calculating the price for multiple years like this helper?

```
/// Public helper to calculate price after a percentage discount has been
/// applied.
public fun calculate_total_after_discount(data: &RequestData, discount: u8): u64 {
    let price = data.base_amount;
    let discount_amount = (((price as u128) * (discount as u128) / 100) as u64);

    price - discount_amount
}
```

Thoralf-M:
> What about calculating the price for multiple years like this helper?
> 
> ```
> /// Public helper to calculate price after a percentage discount has been
> /// applied.
> public fun calculate_total_after_discount(data: &RequestData, discount: u8): u64 {
>     let price = data.base_amount;
>     let discount_amount = (((price as u128) * (discount as u128) / 100) as u64);
> 
>     price - discount_amount
> }
> ```

Not sure what you mean, do you just want to have this part refactored in a helper function or what?
```move
    // Calculate price for first year using registration pricing
    let registration_price = iota_names.get_config<PricingConfig>().calculate_base_price_of_name(name);
    
    // Calculate price for additional years using renewal pricing
    let total_price = if (years == 1) {
        registration_price
    } else {
        let renewal_price = iota_names
            .get_config<RenewalConfig>()
            .config()
            .calculate_base_price_of_name(name);
        
        registration_price + (renewal_price * ((years - 1) as u64))
    };
    ```

**Comment by DaughterOfMars** [comment]
Yes, exactly. Otherwise you won't know how much iota to pay when using a coupon for multiple years

**Comment by Thoralf-M** [comment]
> Yes, exactly. Otherwise you won't know how much iota to pay when using a coupon for multiple years

But you can get the `base_amount` from the resulting PaymentIntent, what's the difference?

**Comment by DaughterOfMars** [comment]
> > Yes, exactly. Otherwise you won't know how much iota to pay when using a coupon for multiple years
> 
> But you can get the `base_amount` from the resulting PaymentIntent, what's the difference?

But then you have to do it as part of the transaction. This would be for checking the cost beforehand.

**Comment by Thoralf-M** [comment]
> But then you have to do it as part of the transaction. This would be for checking the cost beforehand.

You can get the value in a dev inspect, wouldn't be different with the other function?

**Comment by DaughterOfMars** [comment]
> > But then you have to do it as part of the transaction. This would be for checking the cost beforehand.
> 
> You can get the value in a dev inspect, wouldn't be different with the other function?

Then why have the other helper? I think the point is to make is easier

File: packages/iota-names/sources/sales/payment.move [hunk 2]
```diff
@@ -148,6 +150,38 @@ public fun init_registration(iota_names: &mut IotaNames, name: String): PaymentI
...
+    let registration_price = iota_names.get_config<PricingConfig>().calculate_base_price_of_name(name);
+    
+    // Calculate price for additional years using renewal pricing
+    let total_price = if (years == 1) {
+        registration_price
+    } else {
+        let renewal_price = iota_names
+            .get_config<RenewalConfig>()
+            .config()
+            .calculate_base_price_of_name(name);
```
marc2332: Why are you using the renewal price for this? Isn't it wrong to assume they are always the same? Why dont you just use the registration price?

Thoralf-M: They are two different configs and could change, in suins for example renewals are cheaper

DaughterOfMars (approved):

packages/iota-names/sources/sales/payment.move
```diff
@@ -148,6 +150,38 @@ public fun init_registration(iota_names: &mut IotaNames, name: String): PaymentI
     })
 }
 
+/// Creates a `PaymentIntent` for registering a new name for multiple years.
+/// This is a hot-potato and can only be consumed in a single transaction.
+public fun init_registration_with_years(iota_names: &mut IotaNames, name: String, years: u8): PaymentIntent {
```
Alex6323, line 155: I think you could just check the `years` arg for `years == 1`, and then early-return with a call to `init_registration` in that case. This method then becomes a bit cleaner under the assumption of `years >= 2` after the assert.

packages/coupons/tests/coupons_tests.move
```diff
@@ -684,6 +684,190 @@ fun init_renewal(
...
+
+    // Test 100% discount on 4-year registration should result in 0 cost
+    test_coupon_multi_year_register(
+        scenario,
+        b"test.iota".to_string(),
+        4,
+        b"100_PERCENT_OFF".to_string(),
+        user(),
+        option::some(0),
+    );
```
Alex6323, line 789: that feels a bit odd. Why would someone register only 4 years with 100% off coupon? I mean except for stupidity. I guess the question I have is should we maybe restrict certain coupons to 1 year registration only?

packages/coupons/tests/coupons_tests.move
```diff
@@ -684,6 +684,190 @@ fun init_renewal(
...
+                &mut intent,
+                &mut iota_names,
+                coupon_code,
+                &clock,
+                scenario.ctx(),
+            );
+        };
+        if (amount.is_some()) {
+            assert!(intent.request_data().base_amount() == amount.get_with_default(0));
+        };
```
Alex6323, line 857: with `get_with_default(0)` can't you just remove the `if`?

packages/coupons/tests/coupons_tests.move
Alex6323: or separate Coupons into Registration Coupons and Renewal Coupons, or is this making things too complicated? But IMHO a 100%off registration coupon (always 1 year) makes sense. It could be stackable with a 1/2/3/4 year renewal coupon.

packages/coupons/tests/coupons_tests.move
```diff
@@ -684,6 +684,190 @@ fun init_renewal(
...
+        coupon_house::apply_coupon(
+            &mut intent,
+            &mut iota_names,
+            coupon_code,
+            &clock,
+            scenario.ctx(),
+        );
+        if (amount.is_some()) {
+            assert!(intent.request_data().base_amount() == amount.get_with_default(0));
+        };
```
Alex6323, line 820: see comment below

packages/coupons/tests/coupons_tests.move
Thoralf-M: That would need to be adjusted in the coupons module then

packages/iota-names/sources/sales/payment.move
Thoralf-M: Don't see a big difference, but changed it
https://github.com/iotaledger/iota-names/pull/769/commits/843e980fed041d427e80c44a5ad323d9c86f6d9e

packages/coupons/tests/coupons_tests.move
Thoralf-M: Don't think so, without the `if` it might not expect the amount to be the default provided one `0`
But I changed it to extract() https://github.com/iotaledger/iota-names/pull/769/commits/838786f46d6d1472d948c6f6eb8898e1506d1a1d, I think that's better than `get_with_default()`

Alex6323 (approved):

PR #770, opened 2025-09-11
feat(indexer): add admin routes to block strings

### LINKED ISSUE #761: Indexer: support blocking names
## Description

Add two new API endpoints to add/remove strings to be blocked, so they are not returned from the other requests anymore.

## Motivation

It should be possible to block offensive names for the dApp, currently this is handled by https://github.com/iotaledger/iota-names/pull/747, but it breaks the pagination there and also requires new deployments after changes, therefore it would be better to handle this on the indexer side.

## Requirements

1. Add a protected endpoint to add strings to block
2. Add a protected endpoint to unblock strings (for the case where a substring was blocked, that might actually be fine used in other words)
3. Use the strings in the existing API endpoints and filter the names with them by substring without breaking the pagination

Thoralf-M:
Add admin routes to block names that include a blocked string from being returned for auctions

Fixes https://github.com/iotaledger/iota-names/issues/761

Tested with the testnet:
```
export IOTA_NAMES_PACKAGE_ADDRESS=0x7fff6e95f385349bec98d17121ab2bfa3e134f2f0b1ccefc270313415f7835ea
export IOTA_NAMES_OBJECT_ID=0x7cab491740d51e0d75b26bf9984e49ba2e32a2d0694cabcee605543ed13c7dec
export IOTA_NAMES_PAYMENTS_PACKAGE_ADDRESS=0x6b1b01f4c72786a893191d5c6e73d3012f7529f86fdee3bc8c163323cee08441
export IOTA_NAMES_REGISTRY_ID=0x2dfc6f6d46ba55217425643a59dc85fe4d8ed273a9f74077bd0ee280dbb4f590
export IOTA_NAMES_REVERSE_REGISTRY_ID=0x3550bcacb793ef8b776264665e7c99fa3d897695ed664656aac693cf9cf9b76b
export IOTA_NAMES_AUCTION_PACKAGE_ADDRESS=0x6f727ea576a00036657fff0ae3a6d7c8171b178bf35112d6b83b2a6272cc5f0d
export IOTA_NAMES_COUPONS_PACKAGE_ADDRESS=0xa7e4e483d79c245470d5eb3c285a4503a78d90a69d36e35e0993012f5c6137ca
export IOTA_NAMES_SUBNAMES_PACKAGE_ADDRESS=0xd06a5607cc762f2352eeeb8c86c7f962558a06c6023c1eec031a41651d898c87
export IOTA_NAMES_TEMP_SUBNAME_PROXY_PACKAGE_ADDRESS=0x7f34c135e55e5b436b3feaad369eabfe5b6d14c0c57544fefb6921db047e8cbc
export IOTA_NAMES_AUCTION_HOUSE_OBJECT_ID=0x2292ea885039babe8c320f19e0b7546ebdef2b2f6cf2be600bf994cdb51e0050

ADMIN_API_KEY=YOUR_ADMIN_API_KEY RUST_LOG="info,iota_data_ingestion_core=warn" cargo run start --node-url=https://api.testnet.iota.cafe --checkpoint-url=https://checkpoints.testnet.iota.cafe

(stopped after a few seconds, modified progress_store to something around 113269229 to skip many empty blocks, then restarted)

All auctions:
http://localhost:3030/auctions
{"names":["ace-iota-nj.iota","ace-iota-usa.iota","mechita.iota","nounny.iota","teardrop.iota","timson.iota","tioariz.iota","vusd.iota"],"page":0,"pageSize":50,"totalItems":8}

blocking "mechita.iota":
curl -X POST http://localhost:3030/admin/blocked-strings \
  -H "Authorization: Bearer YOUR_ADMIN_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{"string":"mechita.iota"}'
{"success":true,"message":"String 'mechita.iota' has been blocked with full match"} 

mechita.iota not showing up anymore:
http://localhost:3030/auctions
{"names":["ace-iota-nj.iota","ace-iota-usa.iota","nounny.iota","teardrop.iota","timson.iota","tioariz.iota","vusd.iota"],"page":0,"pageSize":50,"totalItems":7}

but it's shown in the blocked strings list:
curl -X GET http://localhost:3030/admin/blocked-strings/detailed \
  -H "Authorization: Bearer YOUR_ADMIN_API_KEY"
{"blockedStrings":["mechita.iota"]} 

unblocking "mechita.iota":
curl -X DELETE http://localhost:3030/admin/blocked-strings \
  -H "Authorization: Bearer YOUR_ADMIN_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{"string":"mechita.iota"}'
{"success":true,"message":"String 'mechita.iota' has been unblocked"}

after unblocking, nothing is blocked anymore:
curl -X GET http://localhost:3030/admin/blocked-strings \
  -H "Authorization: Bearer YOUR_ADMIN_API_KEY"
{"blockedStrings":[]}

also shows up in auctions again:
http://localhost:3030/auctions
{"names":["ace-iota-nj.iota","ace-iota-usa.iota","mechita.iota","nounny.iota","teardrop.iota","timson.iota","tioariz.iota","vusd.iota"],"page":0,"pageSize":50,"totalItems":8}

blocking substring "ace":
curl -X POST http://localhost:3030/admin/blocked-strings \
  -H "Authorization: Bearer YOUR_ADMIN_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{"string":"ace","matchType":"substring"}'
{"success":true,"message":"String 'ace' has been blocked with substring match"}  

ace-iota-nj.iota and ace-iota-usa.iota not showing up anymore:
http://localhost:3030/auctions
{"names":["mechita.iota","nounny.iota","teardrop.iota","timson.iota","tioariz.iota","vusd.iota"],"page":0,"pageSize":50,"totalItems":6}

detailed blocked list after blocking "mechita.iota" also again:
curl -X GET http://localhost:3030/admin/blocked-strings/detailed \
  -H "Authorization: Bearer YOUR_ADMIN_API_KEY"
{"blockedStrings":[{"id":2,"blockedString":"ace","matchType":"substring"},{"id":3,"blockedString":"mechita.iota","matchType":"full"}]} 

pagination works fine with the blocked names removed:
http://localhost:3030/auctions?pageSize=3&page=0
{"names":["nounny.iota","teardrop.iota","timson.iota"],"page":0,"pageSize":3,"totalItems":5}
http://localhost:3030/auctions?pageSize=1&page=1
{"names":["tioariz.iota","vusd.iota"],"page":1,"pageSize":3,"totalItems":5}

```


indexer/src/db/queries.rs
```diff
@@ -133,7 +133,26 @@ pub fn get_auctions_for_bidder(
...
+        .load::<String>(conn)?
+        .into_iter()
+        .filter(|name| {
+            // Filter out names that contain any blocked string
+            !blocked_strings.iter().any(|blocked| name.contains(blocked))
+        })
+        .collect::<Vec<_>>();
+
+    // Apply pagination after filtering
+    let start = page.unwrap_or_default() * page_size;
```
DaughterOfMars: Not sure about this. Maybe it would be better to have a table or a field in the names table that can be used to mark names blocked. Then we can update the table when the admin API calls are made and here we just filter out blocked names in the query. Thoughts?

Thoralf-M: > a table or a field in the names table that can be used to mark names blocked

There is a table now for the blocked strings, what other table do you mean? Regarding a field for a name this would only work for existing entries (unless there is the list of blocked strings + the entry to check on insert if it should be blocked)

But in any case you still need to filter and do the pagination afterwards, or do I miss something?

DaughterOfMars: You would go through the list of new blocked names and update the table of existing names when the admin route is called. Then you can just add a new filter to the query and so it won't affect the user-facing queries.

Thoralf-M: So one list with blocked strings and two lists with names, one of them for the names that are fine and one for the blocked ones?
Then based on a new tx and the blocked strings insert in the correct one
And if the admin blocks/unblocks a string iterate over the lists and move the entries to the other one if needed?

DaughterOfMars: I would just have a new field in the names table that indicates blocked or not.

indexer/src/api/auth.rs
```diff
@@ -0,0 +1,34 @@
...
+    async fn from_request_parts(parts: &mut Parts, state: &S) -> Result<Self, Self::Rejection> {
+        const ADMIN_API_KEY_ENV: &str = "ADMIN_API_KEY";
+
+        let TypedHeader(Authorization(bearer)) =
+            TypedHeader::<Authorization<Bearer>>::from_request_parts(parts, state)
+                .await
+                .map_err(|_| ApiError::BadRequest("Missing authorization header".to_string()))?;
+
+        let expected_key = std::env::var(ADMIN_API_KEY_ENV)
+            .map_err(|_| ApiError::Unauthorized("Admin API key not configured".to_string()))?;
```
Alex6323, line 26: not sure this is the right place to do this as it would read it over and over from the env. We probably should read at the start of the indexer and store the expected key in the state :thinking:

indexer/src/api/responses.rs
```diff
@@ -17,3 +17,40 @@ impl axum::response::IntoResponse for AuctionsResponse {
...
+#[serde(rename_all = "camelCase")]
+pub struct BlockStringRequest {
+    pub string: String,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct UnblockStringRequest {
+    pub string: String,
+}
```
Alex6323: bit weird that those requests are in a "responses" module. Maybe just call the module "types" or create a separate "requests" module?

indexer/src/db/queries.rs
```diff
@@ -133,7 +133,26 @@ pub fn get_auctions_for_bidder(
         SortOrder::Desc => query.order(names::name.desc()),
     };
 
-    Ok(query.load(conn)?)
+    let mut results = query
+        .load::<String>(conn)?
+        .into_iter()
+        .filter(|name| {
+            // Filter out names that contain any blocked string
+            !blocked_strings.iter().any(|blocked| name.contains(blocked))
```
Alex6323: don't see how this will work yet. Imagine there a 2 auctions: `analysis.iota` and `anal.iota`. Admin sees the latter and uses his admin superpowers to block it. Now the other one from an honorable mathematical institute gets their auction also blocked?

Alex6323: maybe what we also need is to let the admin provide a ruleset (bitmask?) associated with the string to block. For instance, some words are not offensive if they contain something offensive as a substring BUT only as a substring, some other strings might always be offensive, whether as a substring or full string. Sometimes something is offensive if it's a prefix/suffix, but not otherwise, or separated with an `_` (which is one of the allowed name chars). 

I think we need to be careful here in order to not shoot ourselves in the foot. Luckily we have a very limited characterset for names, so something like this could be feasible (compared to full blown regexes)

Alex6323: Also I agree with Chloé if we can mark names as offensive on insert and on admin call only, then this should not even be that costly.

indexer/src/api/auth.rs
Thoralf-M: Tbh I think it's fine like this, the overhead to load a simple env value is very small and only needed for admin calls which are also rare
And this way one can set/change the env after the server is started without a restart

Alex6323: good points

indexer/src/db/queries.rs
Thoralf-M: Changed in https://github.com/iotaledger/iota-names/pull/770/commits/048d33e589a501b9ab9d603e543cfc8c3efcfe56

indexer/src/db/queries.rs
Thoralf-M: Good point with the false positives, I changed it so one can now specify if the full string should be blocked or also substring already for cases like `verybadword1`, `verybadword2`, `verybadword3`

Thoralf-M: Also modified to filter on insertion https://github.com/iotaledger/iota-names/pull/770/commits/048d33e589a501b9ab9d603e543cfc8c3efcfe56

Thoralf-M:
Will be handled in a different way

PR #775, opened 2025-09-24
fix(dapp): fix infinite render loop error

### LINKED ISSUE #764: fix infinite render loop
This error happens when entering the My Names page 

Image

Image

KeitoTadashi:
fixes #764

### DIFF
--- a/dapp/src/components/name-record/AvatarDisplay.tsx
+++ b/dapp/src/components/name-record/AvatarDisplay.tsx
@@ -20,10 +20,9 @@ interface NameAvatarDisplay {
 
 export function NameAvatarDisplay({ name }: NameAvatarDisplay) {
     const { data: nameRecordData, isLoading: isNameRecordDataLoading } = useNameRecord(name);
-    const isNameSubname = isSubname(name);
-    // load avatar for names
+
     const avatarId =
-        !isNameSubname && nameRecordData?.type === 'unavailable'
+        nameRecordData?.type === 'unavailable'
             ? nameRecordData?.nameRecord.avatar
                 ? nameRecordData.nameRecord.avatar
                 : null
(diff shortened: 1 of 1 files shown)


dapp/src/components/name-record/AvatarDisplay.tsx
```diff
@@ -19,24 +18,20 @@ interface NameAvatarDisplay {
...
     const { data: nameRecordData, isLoading: isNameRecordDataLoading } = useNameRecord(name);
-    const { data: subnames, isLoading: isSubnamesLoading } = useRegistrationNfts('subname');
-    const isNameSubname = isSubname(name);
 
     const avatarId =
         nameRecordData?.type === 'unavailable'
             ? nameRecordData?.nameRecord.avatar
                 ? nameRecordData.nameRecord.avatar
-                : isNameSubname
-                  ? subnames?.find((n) => n.name === name)?.id
```
msarcev, line 30: That's there for a reason, no ?
Subname avatars are not loading if you remove this.

msarcev (changes requested):

dapp/src/components/name-record/AvatarDisplay.tsx
marc2332: Yeah lol, this is removing an actual feature

KeitoTadashi: with latest changes we fetch all avatar for all types of names/subnames without infinite render loop @msarcev @marc2332

KeitoTadashi: Let me explain myself a bit more:

Infinite loop rendering was caused by the hook call [useRegisterNft](https://github.com/iotaledger/iota-names/pull/775/files#diff-44bc159bfdcc8b952e2ce74ee20875aa458ce2864d248a6222daa283b2faa3e2L22)

For every avatar we try to get, we fetch all subnames as a result, +1000 hook calls were made across dapp when is trying to fetch avatars.

The source of this problem is the well-know problem with [subnames nft id](https://boxfishstudio.slack.com/archives/C08EGKXAMN2/p1750158709395809)

There are only 2 types of <NameAvatarDisplay>
-> Avatar set by user, which is stored in [avatar property](https://github.com/iotaledger/iota-names/pull/775/files#diff-44bc159bfdcc8b952e2ce74ee20875aa458ce2864d248a6222daa283b2faa3e2R26)
-> Avatar created by us with [names-display API](https://github.com/iotaledger/iota-names/pull/775/files#diff-44bc159bfdcc8b952e2ce74ee20875aa458ce2864d248a6222daa283b2faa3e2R38)

If avatar is set by user, then we will access the address of it from the property of RegistrationNft Object, and after that we get ObjectId of that AvatarId
But if the avatar is not set by user and it's a subname, then we fetch parent ID and with it, we try to get Object ID from parent. (the weird thing here is that instead of parent's avatar, it will show the names-display avatar)

So my solution is: 
If avatar from registrationNft property is null/empty, then we show the names-display avatar, without making any hook call

@marc2332 @msarcev

msarcev (approved):

marc2332 (approved):

PR #785, opened 2025-09-29
feat(packages): Add lookup methods that check expiration

### LINKED ISSUE #768: move: Add verified lookup and reverse_lookup functions
## Description

Add `public fun lookup_address(self: &Registry, name: Name, clock: &Clock): Option<Address> {` and `public fun reverse_lookup(self: &Registry, address: address, clock: &Clock): Option<Name> {`, maybe with `_verified` suffix to distinguish with the functions in registry.move.

## Motivation

Currently, callers need to manually call the checks if they want to be sure that something is not expired:
```move
    /// Different custom error messages.
    #[error]
    const ENameNotFound: vector<u8> = b"Name not found.";
    #[error]
    const ENameNotPointingToAddress: vector<u8> = b"Name not pointing to an address.";
    #[error]
    const ENameExpired: vector<u8> = b"Name expired.";
 
    /// A function to transfer an object of any type T to a name (for instance `example.iota`)
    public fun send_to_name<T: key + store>(iota_names: &IotaNames, obj: T, name: String, clock: &Clock) {
        // Look up the name on the registry.
        let mut optional = iota_names.registry<Registry>().lookup(name::new(name));
        // Check that the name indeed exists.
        assert!(optional.is_some(), ENameNotFound);
 
        let name_record = optional.extract();
        // Check that name has not expired. 
        // This check is optional, but it's recommended you perform the verification.
        assert!(!name_record.has_expired(clock), ENameExpired);
        // Check that the name has a target address set.
        assert!(name_record.target_address().is_some(), ENameNotPointingToAddress);
 
        // Transfer the object to that name.
        transfer::public_transfer(obj, name_record.target_address().extract())
    }
```
Having high level methods that do these checks internally makes it easier to use without doing something wrong

## Requirements

Add the two new functions

## Open questions (optional)

Are other names better? Should they return an Option or just error instead?

DaughterOfMars:
## Description

Adds the `lookup_unexpired`, `lookup_address`, `resolve_address`, and `reverse_lookup_unexpired` functions to the `iota-names::registry` module, which check if the record is expired before returning anything.

Closes #768

### DIFF
--- a/packages/iota-names/sources/registry.move
+++ b/packages/iota-names/sources/registry.move
@@ -342,10 +342,41 @@ public fun lookup(self: &Registry, name: Name): Option<NameRecord> {
     }
 }
 
+/// Returns the `NameRecord` associated with the given name and ensures that it is not expired,
+/// or None otherwise.
+public fun lookup_verified(self: &Registry, name: Name, clock: &Clock): Option<NameRecord> {
+    if (self.registry.contains(name)) {
+        let record = &self.registry[name];
+
+        // Check that name has not expired.
+        assert!(!record.has_expired(clock), ERecordExpired);
+
+        some(*record)
+    } else {
+        none()
+    }
+}
+
 /// Returns the `name` associated with the given address or None.
 public fun reverse_lookup(self: &Registry, address: address): Option<Name> {
     if (self.reverse_registry.contains(address)) {
         some(self.reverse_registry[address])
+    } else {
+        none()
+    }
+}
+
+/// Returns the `name` associated with the given address and ensures it is not expired,
+/// or None otherwise.
+public fun reverse_lookup_verified(self: &Registry, address: address, clock: &Clock): Option<Name> {
+    if (self.reverse_registry.contains(address)) {
+        let name = self.reverse_registry[address];
+        let record = &self.registry[name];
+
+        // Check that name has not expired.
+        assert!(!record.has_expired(clock), ERecordExpired);
+
+        some(name)
     } else {
         none()
     }
--- a/packages/iota-names/sources/registry.move
+++ b/packages/iota-names/sources/registry.move
@@ -366,10 +366,12 @@ public fun lookup_address(self: &Registry, name: Name, clock: &Clock): Option<ad
     }
 }
 
-/// Returns the address associated with the given name if one is set and the record is not expired,
-/// or None otherwise. This function is callable from PTBs unlike `lookup_address`.
-public fun resolve_address(iota_names: &IotaNames, name: Name, clock: &Clock): Option<address> {
-    iota_names.registry<Registry>().lookup_address(name, clock)
+/// Returns the address associated with the given name if one is set and the record is not expired. 
+/// This function is callable from PTBs unlike `lookup_address`.
+public fun resolve_address(iota_names: &IotaNames, name: Name, clock: &Clock): address {
+    let mut address = iota_names.registry<Registry>().lookup_address(name, clock);
+    assert!(address.is_some(), ERecordNotFound);
+    address.extract()
 }
 
 /// Returns the `name` associated with the given address or None.
--- a/packages/iota-names/sources/registry.move
+++ b/packages/iota-names/sources/registry.move
@@ -345,13 +345,24 @@ public fun lookup(self: &Registry, name: Name): Option<NameRecord> {
 /// Returns the `NameRecord` associated with the given name and ensures that it is not expired,
 /// or None otherwise.
 public fun lookup_verified(self: &Registry, name: Name, clock: &Clock): Option<NameRecord> {
+    let record = self.lookup(name);
+    if (record.is_some()) {
+        // Check that name has not expired.
+        assert!(!record.borrow().has_expired(clock), ERecordExpired);
+    };
+    record
+}
+
+/// Returns the address associated with the given name if one is set and the record is not expired,
+/// or None otherwise.
+public fun lookup_address(self: &Registry, name: Name, clock: &Clock): Option<address> {
     if (self.registry.contains(name)) {
         let record = &self.registry[name];
 
         // Check that name has not expired.
         assert!(!record.has_expired(clock), ERecordExpired);
 
-        some(*record)
+        record.target_address()
     } else {
         none()
     }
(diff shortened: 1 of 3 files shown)


Thoralf-M (changes requested):
Please also add tests for them

packages/iota-names/sources/registry.move
```diff
@@ -342,6 +342,21 @@ public fun lookup(self: &Registry, name: Name): Option<NameRecord> {
     }
 }
 
+/// Returns the `NameRecord` associated with the given name and ensures that it is not expired,
+/// or None otherwise.
+public fun lookup_verified(self: &Registry, name: Name, clock: &Clock): Option<NameRecord> {
```
Thoralf-M: Wondering if we want the Option here or if we want two functions, one with Option so one can handle the None case and one without
Because in many cases one would probably want to error anyways and then with the Option you have to manually call .extract() or something similar on it which makes it less convenient
Or maybe even just a function to get the address directly, not even the NameRecord. Because this seems to be the main use case, provide the name and get the address so you can send funds there and just error if there is no address set

DaughterOfMars: I think the `Option` is fine, as otherwise we'll have one fn that returns `Option` and one that doesn't and that's a bit weird. Also, using errors instead means that there is no way for users to use this without failing, like to check if a name exists for instance.

Alex6323: Do we want to error at all? The record not existing or being expired seem similar scenarios, and the caller's logic might take that into account. I am imagining an iteration over some names and picking the first name that has a "verified" (existing + not expired) `NameRecord`. So I feel like we should just have it return an `Option`, but also return `None` when it's expired. Or is the abort explicitly wanted?

Thoralf-M: Maybe we need even more fns for the different use cases
What I was thinking of mainly is the case where you would just call `lookup_address()` as first thing in a PTB (could even be a fn on the transaction builder that accepts the name) and then use the result as address to send funds
In this case you don't want to deal with options and handle potential none cases, you just want to send or fail

So one can just add something similar to this
```Rust
    let iota_names_object_id =
        ObjectId::from_str("0x7cab491740d51e0d75b26bf9984e49ba2e32a2d0694cabcee605543ed13c7dec")?;
    builder
        .move_call(
            "0x7fff6e95f385349bec98d17121ab2bfa3e134f2f0b1ccefc270313415f7835ea".parse()?,
            "registry",
            "lookup_address",
        )
        .arguments(vec![
            Input::Shared {
                object_id: iota_names_object_id,
                initial_shared_version: 512290735,
                mutable: false,
            },
            Input::Pure {
                value: "name.iota".as_bytes().to_vec(),
            },
        ])
        .name("name_address");
```

packages/iota-names/sources/registry.move
```diff
@@ -342,6 +342,32 @@ public fun lookup(self: &Registry, name: Name): Option<NameRecord> {
...
+    if (record.is_some()) {
+        // Check that name has not expired.
+        assert!(!record.borrow().has_expired(clock), ERecordExpired);
+    };
+    record
+}
+
+/// Returns the address associated with the given name if one is set and the record is not expired,
+/// or None otherwise.
+public fun lookup_address(self: &Registry, name: Name, clock: &Clock): Option<address> {
```
Thoralf-M, line 361: We should have this also with IotaNames object as parameter and get the registry from it to call this function internally, otherwise one can't call it from a PTB and also in smart contracts it's one more step, because you need to get the Registry first

packages/iota-names/sources/registry.move
Alex6323: okay, I see, but now that I see your example,  I think we really need to have a high-level function on the transaction builder that takes a name, so a `send_coins_to_name()`, `transfer_objects_to_name()`, etc methods, which does the lookup internally.

DaughterOfMars: I reworked the methods, let me know what you think

packages/iota-names/sources/registry.move
DaughterOfMars: done

packages/iota-names/sources/registry.move
```diff
@@ -332,16 +332,46 @@ public fun has_record(self: &Registry, name: Name): bool {
...
+    if (self.has_unexpired_record(name, clock)) {
+        self.registry[name].target_address()
+    } else {
+        none()
+    }
+}
+
+/// Returns the address associated with the given name if one is set and the record is not expired,
+/// or None otherwise. This function is callable from PTBs unlike `lookup_address`.
+public fun resolve_address(iota_names: &IotaNames, name: Name, clock: &Clock): Option<address> {
```
Thoralf-M: Not sure about the `_extracted`, but I would also add another function like this:
```move
/// Returns the address associated with the given name if one is set and the record is not expired,
/// panics otherwise. This function is callable from PTBs unlike `lookup_address`.
public fun resolve_address_extracted(iota_names: &IotaNames, name: Name, clock: &Clock): address {
    iota_names.registry<Registry>().lookup_address(name, clock).extract()
}
```

DaughterOfMars: but why

Thoralf-M: So you get the address with a single move call, if you want to use it in PTBs it's much simpler because you don't have to call the extract with the correct type

DaughterOfMars: Hmm, I see. But really, how much do we want to add just for PTBs?

DaughterOfMars: Maybe instead we just make the `resolve_address` fn extract it since it's already there for PTBs

Thoralf-M: I don't think a few more fns will hurt and they just make it easier for users with specific needs. Personally I think the address without option is more important regarding that, but I think both are good to have

DaughterOfMars: I think it makes it kind of confusing which one to call, so for now I prefer to have just one extra method

Thoralf-M (approved):

Alex6323 (approved):

PR #786, opened 2025-09-29
chore(packages): Improve error message when new subname does not match parent

### LINKED ISSUE #778: Return better error in iota_names_subnames::subnames::new_leaf
## Description

using iota_names_subnames::subnames::new_leaf the function should just take the leaf name and not the full name since the parent name is already one of the parameters. If this is not wanted then at least return a proper error that the subname does not match the parent

DaughterOfMars:
## Description

Improves the error message for when a new subname is created but does not match the parent, i.e. passing `test.subname.iota` with parent `sub.iota`.

Closes #778

### DIFF
--- a/packages/subnames/sources/config.move
+++ b/packages/subnames/sources/config.move
@@ -64,10 +64,10 @@ public fun new(
 
 /// Validates that the child name is a valid child for parent.
 public fun assert_is_valid_subname(parent: &Name, child: &Name, config: &SubnameConfig) {
+    assert!(is_parent_of(parent, child), EInvalidParent);
     assert!(is_valid_tln(child, config), ENotSupportedTLN);
     assert!(is_valid_label(child, config), EInvalidLabelSize);
     assert!(has_valid_depth(child, config), EDepthExceedsLimit);
-    assert!(is_parent_of(parent, child), EInvalidParent);
 }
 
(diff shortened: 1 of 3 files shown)


Thoralf-M (approved):

packages/subnames/sources/config.move
```diff
@@ -64,10 +64,10 @@ public fun new(
 
 /// Validates that the child name is a valid child for parent.
 public fun assert_is_valid_subname(parent: &Name, child: &Name, config: &SubnameConfig) {
-    assert!(is_valid_tln(child, config), ENotSupportedTLN);
     assert!(is_valid_label(child, config), EInvalidLabelSize);
-    assert!(has_valid_depth(child, config), EDepthExceedsLimit);
     assert!(is_parent_of(parent, child), EInvalidParent);
+    assert!(is_valid_tln(child, config), ENotSupportedTLN);
```
Alex6323, line 69: why is the reordering necessary? checking that the TLN is valid first seems like a good first check, no?

DaughterOfMars: I think the point is there's no perfect order but this way if you just pass one label you'll get a more useful error

Alex6323 (approved):

PR #788, opened 2025-09-30
feat(dapp): add cookie disclaimer

### LINKED ISSUE #795: Add cookie disclaimer to iota-names dapp

Juligs:


KeitoTadashi (approved):
🚀

dapp/src/components/disclaimer/CookieDisclaimer.tsx
```diff
@@ -0,0 +1,51 @@
...
+    consentToAnalytics,
+    declineAnalytics,
+    initAnalytics,
+} from '@/lib/utils/analytics/amplitude';
+
+const defaultNetwork = CONFIG.network;
+
+export function CookieDisclaimer() {
+    useEffect(() => {
+        initAnalytics(defaultNetwork);
```
marc2332, line 22: Shouldn't you make it only track once approved?

marc2332: Or is this temporary?

marc2332 (changes requested):

dapp/src/components/disclaimer/CookieDisclaimer.tsx
Juligs: If we initialize Amplitude only after the user clicks “accept,” we lose the tracking data from before they accept the cookies. What it actually does is initialize right away, but it doesn’t send any events until the user gives consent.

```
  await ampli.load({
        environment: 'iotanames',
        disabled: !IS_PROD_ENV,
        client: {
            configuration: {
                optOut: true, // Start with tracking disabled for GDPR compliance
                autocapture: true,
            },
        },
    });
```

panteleymonchuk (approved):

dapp/src/providers/AppProviders.tsx
```diff
@@ -23,11 +24,6 @@ export function AppProviders({ children }: React.PropsWithChildren) {
     const allNetworks = getAllNetworks();
     const defaultNetwork = CONFIG.network;
 
-    useEffect(() => {
```
msarcev, line 26: Why are we removing this ?

msarcev: This is the init currently commented until we decide to enable it.

msarcev (changes requested):

dapp/src/providers/AppProviders.tsx
Juligs: I didn’t delete it, I just moved it to `CookieDisclaimer`

 ```
export function CookieDisclaimer() {
    useEffect(() => {
        initAnalytics(defaultNetwork);
    }, []);
```

**Review by msarcev — APPROVED** [review]

**Review by marc2332 — CHANGES_REQUESTED** [review]
if I reject and reload, it asks me again, seems like the cookie is not being set to `false`

File: dapp/src/components/dialogs/PrivacyPolicyDialog.tsx [hunk 3]
```diff
@@ -0,0 +1,102 @@
...
+import { useState } from 'react';
+
+export function PrivacyPolicyDialog({ configuration }: { configuration: SKCMConfiguration }) {
+    const router = useRouter();
+    const searchParams = useSearchParams();
+    const open = searchParams.get('modal') === 'privacy_policy';
+
+    const { onAcceptCookies, onDeclineCookies } = configuration;
+    const submitNecessaryCookies = useSubmitNecessaryCookies();
+    const [loading, setLoading] = useState(false);
```
marc2332: What is this state for? It never appears "loading", and looking at the code I dont see why would it either

marc2332 (changes requested):

dapp/src/components/dialogs/PrivacyPolicyDialog.tsx
marc2332: Imo it can be removed entirely

marc2332 (approved):

PR #803, opened 2025-10-15
fix(tests): add typecheck to typescript CI and fix errors

### LINKED ISSUE #554: Add typecheck to TS tests
TS tests have type errors introduced by latest PR merges. Check `localnet.test.ts` and make the CI fail if there are any errors in any test file

Image

VmMad:
Fixes #554


sdk/tsconfig.lint.json
marc2332, line 1: Why do we need another tsconfig? Cant you simply use `tsconfig.json` and add `test` in there?

marc2332 (changes requested):

sdk/tsconfig.lint.json
VmMad: for that to work i'd have to change the tsconfig.json's `rootDir` which changes the way the types are generated into the `dist` folder. The easiest way is creating another specific tsconfig that works for both folders

marc2332 (approved):

brancoder (approved):
LGTM

PR #831, opened 2025-10-28
feat: upgrade devnet `iota-names` and `subnames` packages

thibault-martinez:


Thoralf-M (dismissed):

scripts/package-info/devnet.json
```diff
@@ -10,13 +10,13 @@
         }
     },
     "iotaNamesObjectId": "0x07c59b37bd7d036bf78fa30561a2ab9f7a970837487656ec29466e817f879342",
-    "packageId": "0xb9d617f24c84826bf660a2f4031951678cc80c264aebc4413459fb2a95ada9ba",
+    "packageId": "0xd8d28a7446a93b1cc0ae248d19b649b9f9b5e0b05332a31aa3ab514e4b7aff32",
```
Thoralf-M: Actually not sure about this change here, since the types still have the old version
We might need to change it so we can have multiple ids

thibault-martinez: Yeah I wasn't sure either 😬

Thoralf-M: Might be best to do it like this
```json
"packageIds": {
    "1": "0xb9d617f24c84826bf660a2f4031951678cc80c264aebc4413459fb2a95ada9ba",
    "2": "0xd8d28a7446a93b1cc0ae248d19b649b9f9b5e0b05332a31aa3ab514e4b7aff32"
}
```
@marc2332 @msarcev opinions? In the dApp you will need to use version 1 for now, but if we do more changes you might need to use different versions for different functions/types
Also note that this version will be different from testnet to mainnet, testnet version 3 might contain the same features than mainnet version 2

alexsporn: Then it would be better to instead configure the package to be be used for a certain operation, even if those are repeated in the config, else you distribute this logic everywhere in the code

KeitoTadashi (approved):
new contracts works correctly, now we are able to [register](https://explorer.iota.org/txblock/AEkFASBkUDZyaxmcFpEV3qV8qPgcWnGPY3AuwEw32ccj?network=devnet) a name with the new function `init_registration_with_years` using devnet

scripts/package-info/devnet.json
thibault-martinez: https://github.com/iotaledger/iota-names/pull/831/commits/656ee37f29723161c5b32850b91b16e94ff8d140

thibault-martinez: https://github.com/iotaledger/iota-names/pull/831/commits/656ee37f29723161c5b32850b91b16e94ff8d140

salaheldinsoliman (approved):

marc2332 (approved):

Thoralf-M (approved):

msarcev (approved):

PR #864, opened 2025-11-10
feat(dapp): update auction action button

### LINKED ISSUE #818: Update finished auction button state
After an auction has finished, users should no longer be able to access the bidding screen: [LINK](https://jam.dev/c/d387fee7-5fb4-4673-b2e6-46b7985ee2cd)

The button should immediately update from “Bid Again” to “Claim” without requiring a page refresh.

KeitoTadashi:
fixes #818


dapp/src/auctions/components/AuctionActionButton.tsx
```diff
@@ -20,6 +21,21 @@ export function AuctionActionButton({
     auctionStatus: UserAuctionStatus;
     onBidClick?: (targetName: string) => void;
 }) {
+    const [, setForceUpdate] = useState(0);
```
msarcev: Is the point just to trigger a re-render ?

KeitoTadashi: yess

marc2332: https://github.com/iotaledger/iota-names/commit/e6378d7bc429b101b72a2f82aec1c2d51ff805c2

marc2332 (approved):

dapp/src/auctions/lib/utils/metadata.ts
```diff
@@ -137,18 +141,26 @@ export function createAuctionMetadataQuery({
...
                 auctionsTableObjectId,
                 packageId,
                 graphQLClient,
             });
+            const auctionMetadata = parseAuctionDataBcs(data);
+            const auction = auctionMetadata?.value.value;
+
+            const now = new Date();
+            const endTimestamp = new Date(Number(auction?.end_timestamp_ms));
+            const isActive = now < endTimestamp;
```
brancoder, line 155: can we move this code  for  isActive flag calculation to normalizeAuctionData function instead of passing this flagg around?

dapp/src/auctions/lib/utils/metadata.ts
```diff
@@ -72,7 +72,10 @@ export async function fetchAuctionMetadata({
 /**
  * Converts raw BCS auction data to normalized TypeScript interface
  */
-function normalizeAuctionData(rawData: RawAuctionMetadata): AuctionMetadata | null {
+function normalizeAuctionData(
+    rawData: RawAuctionMetadata,
+    isActive: boolean,
+): AuctionMetadata | null {
```
brancoder, line 78: instead of accepting the isActive flage we can calculate it inside of the function ex.
```
const endTimestamp = new Date(Number(auction.end_timestamp_ms));
    const isActive = new Date() < endTimestamp;
```

brancoder (changes requested):
looks good and works nice just a comment about flag calculation

dapp/src/auctions/lib/utils/metadata.ts
KeitoTadashi: no, we can not

dapp/src/auctions/lib/utils/metadata.ts
KeitoTadashi: not updated flag

dapp/src/auctions/lib/utils/metadata.ts
brancoder: tested the structural sharing flag to false and it works. 
the issue here is that isActive depends on current time, not just on-chain data. so setting the structural sharing to false causes the select to re run and than the isActive calculation can be moved to normalize function.
add this to `createAuctionMetadataQuery` and revert all other changes.
      `structuralSharing: false,`
      
and than do the `isActive` calculation in the normalize function

brancoder (changes requested):

dapp/src/auctions/lib/utils/metadata.ts
brancoder: we just need to add `structuralSharing: false,` to the query and it will work

dapp/src/auctions/lib/utils/metadata.ts
marc2332: No, `isActive`must be in query fn so that if it changes the query is marked as updated.

marc2332 (changes requested):
The latest commits have undone my fix, so the issue is back

explanation:

When useAuctionRefresh calls invalidateQueries the queryFn reruns, if the returned value is the same then nothing happens because tanstack sees that the data is 100% the same

so the UI doesnt rerun

which is precisely the issue that the PR is meant to fix
the fact that the UI is not refreshed

so I moved isActive into the queryFn so that when it reruns (trigerred by invalidateQueries) tanstack sees that the returned data is different (isActive went from true to false) making it notify the components and thus rerunning the UI
and ultimately "refreshing" it

dapp/src/auctions/lib/utils/metadata.ts
```diff
@@ -137,20 +141,23 @@ export function createAuctionMetadataQuery({
...
+            data: { auctionMetadata: RawAuctionMetadata | null } | null,
         ): AuctionMetadata | null => {
-            const auctionMetadataBcs = parseAuctionDataBcs(data);
-            return normalizeAuctionData(auctionMetadataBcs);
+            if (!data?.auctionMetadata?.value?.value) return null;
+            return normalizeAuctionData(data.auctionMetadata);
         },
         enabled: !!auctionsTableObjectId && !!name,
         refetchInterval: 10000,
+        structuralSharing: false, //This is necessary because isActive depends on time, not just on-chain data
```
marc2332: Not needed if done like it was done before

brancoder (approved):
LGTM

marc2332 (approved):

PR #953, opened 2025-11-21
feat(dapp): Set Avatar NFT test

### LINKED ISSUE #886: Name management tests: Set NFT

KeitoTadashi:
fixes #886


msarcev (approved):

marc2332 (changes requested):
We can't use the names their selves as assets

marc2332 (approved):

PR #970, opened 2025-11-25
chore(dapp): Remove TX dryRuns from e2e tests

msarcev:
Dry runs are unnecessary work in e2e tests. We are doing the TX after the dryRun in any case, and expect the TX to work. The test will notify us if the TX fails


KeitoTadashi (approved):
sublime

VmMad (approved):

PR #971, opened 2025-11-25
feat(indexer): add reset metrics flag

### LINKED ISSUE #906: Add indexer flag to allow starting metrics from 0
Add a new indexer CLI flag so it becomes explicit we want to start metrics from 0. It would refuse to start if we start without metrics and without the flag. It would avoid accidentally starting from scratch.

Alex6323:
Closes https://github.com/iotaledger/iota-names/issues/906


indexer/src/main.rs
```diff
@@ -60,6 +60,9 @@ enum Command {
         /// The URL of Prometheus to restore metrics from on startup.
         #[arg(long, default_value = "http://localhost:9090")]
         prometheus_url: String,
+        /// Resets metrics in case of a Prometheus error.
+        #[arg(long, default_value_t = false)]
+        reset_metrics_on_error: bool,
```
Thoralf-M: Please mention this option in the README.md https://github.com/iotaledger/iota-names/tree/develop/indexer#start-the-indexer

Thoralf-M (approved):

indexer/src/main.rs
thibault-martinez: Let's simply call it `reset_metrics` please. Even though the trigger is a prometheus error, this flag shows the intent of the user.

Alex6323: https://github.com/iotaledger/iota-names/pull/971/commits/08f11cc620501e2fa04e4b4bebbec0ebc6e497d2

thibault-martinez (approved):

PR #972, opened 2025-11-25
feat(dapp): Polish dropdowns

### LINKED ISSUE #901: Polish expiration dropdown
See https://github.com/iotaledger/if-comms/issues/2110 for reference
Lets go for this solution (use IOTA, not fiat)

Image

We should apply this to:
- direct purchases after #636 
- renew

We should also NOT compute the gas in the price list, the price list should reflect just the pice, not the extra gas

KeitoTadashi:
fixes #901


dapp/src/components/dialogs/RenewNameDialog.tsx
```diff
@@ -101,9 +100,7 @@ export function RenewNameDialog({ setOpen, name, onRenew }: RenewDialogProps) {
     const [coupons, setCoupons] = useState<UserSetCoupon[]>([]);
     const [applyCoupons, setApplyCoupons] = useState(false);
 
-    const { data: renewalPriceInNanos } = useCalculateRenewalPrice(name, renewYears ?? 1);
-    const renewalPriceIota = renewalPriceInNanos ? formatNanosToIota(renewalPriceInNanos) : '0';
-    const fiatPriceResult = useCalculatePriceInFiat(renewalPriceInNanos || '0');
+    // const fiatPriceResult = useCalculatePriceInFiat(renewalPriceInNanos || '0');
```
marc2332: debris?

marc2332 (changes requested):

dapp/src/components/dialogs/RenewNameDialog.tsx
KeitoTadashi: i leave it here on purpose because probably we will add it later somewhere in the dialog

msarcev: God forbid we lose that line, who would be able to recover it. :trollface:

marc2332 (approved):
LGTM, but let's remove those commented lines we dont need them in the code atm so they can go away

marc2332 (changes requested):
Shouldn't it be more rounded like ?:

Image


Yours:

Image

Also a slightly different bg color

msarcev (approved):

dapp/src/components/dialogs/PurchaseNameDialog.tsx
```diff
@@ -268,11 +268,40 @@ export function PurchaseNameDialog({ name, open, setOpen, onPurchase }: Purchase
...
-    const purchaseOptions: SelectOption[] = Array.from({ length: purchaseableYears }, (_, i) => ({
-        id: String(i + 1),
-        label: `${i + 1} Year${i ? 's' : ''}`,
-    }));
+    const purchaseYearsArray = Array.from({ length: purchaseableYears }, (_, i) => i + 1);
+
+    const purchasePricesQueries = useQueries({
+        queries: purchaseYearsArray.map((year) => ({
+            queryKey: [...queryKey.renewalPrice(name, year)],
+            queryFn: async () => {
```
panteleymonchuk: nit: not sure you need async/await here

dapp/src/components/dialogs/RenewNameDialog.tsx
```diff
@@ -193,10 +184,40 @@ export function RenewNameDialog({ setOpen, name, onRenew }: RenewDialogProps) {
             : 0;
 
     const isRenewable = (renewableYears ?? 0) > 0;
-    const renewOptions: SelectOption[] = Array.from({ length: renewableYears }, (_, i) => ({
-        id: String(i + 1),
-        label: `${i + 1} Year${i ? 's' : ''}`,
-    }));
+    const renewYearsArray = Array.from({ length: renewableYears }, (_, i) => i + 1);
```
panteleymonchuk: The only difference between logic from PurchaseNameDialog and RenewNameDialog is `renewableYears|purchaseableYears` values.
Logic about `...YearsArray`, `pricesQueries`, `purchaseOptions` the same.

Could we join it somehow in the component/hook, so we don't repeat ourself?

panteleymonchuk (changes requested):

panteleymonchuk (approved):

marc2332 (approved):

PR #979, opened 2025-11-26
feat(dapp): add set avatar amplitude event

### LINKED ISSUE #960: Add set avatar amplitude event
Trigger: User sets/unsets an NFT as an avatar for one of his domains
Properties: 
 - name
 - setAvatar → true/false

Juligs:
fixes #960


msarcev (approved):

panteleymonchuk (approved):
works for me

PR #993, opened 2025-12-02
chore(dapp): polish existing amplitude events

### LINKED ISSUE #958: Polish existing amplitude events
- Start/Bid Auction --> add `coinType` (in case if we add other currencies in the future)
- Purchase name --> add `coinType` (in case if we add other currencies in the future)
- Renew Subname --> 
    - `expiration` → distinguish between `parent` and `custom`
    - `expirationDate` → the real rexpiration time
- Create subname --> update properties so we have 2 event names: `subname` & `parentName` where the parent is the whole parent, which could be a subdame in itself like subname: lala, parent: lala.lala.tooling
- Combine connected address and set address as displayed in 1 which is connected address: properties: 
    - `name`
    - `addressType` → current / external
    - `setNameAsDisplayed` → true / false

See https://iotafoundation.atlassian.net/wiki/spaces/IOTANames/pages/2621931526/IOTA+Names+-+Amplitude+Events for more details (red is whats missing)

panteleymonchuk:
Closes #958


marc2332 (approved):

msarcev (dismissed):

msarcev (changes requested):
Needs a test run

msarcev (changes requested):
Conflicts

msarcev (approved):

PR #1011, opened 2025-12-11
feat(typescript-ci): Add script to setup network for e2e tests

### LINKED ISSUE #996: Improve e2e CI run to inject iota-names configs in the node
Necessary for `lookup` and `reverse-lookup` related tests to work.

msarcev:
Closes https://github.com/iotaledger/iota-names/issues/996

Needs https://github.com/iotaledger/iota/pull/9538 to be in the binaries we download


KeitoTadashi (approved):

.github/scripts/setup-e2e-network.sh
```diff
@@ -0,0 +1,313 @@
...
+max-type-argument-depth = 16
+max-type-argument-width = 32
+max-type-nodes = 256
+max-move-value-depth = 128
+max-transaction-ids = 1000
+max-scan-limit = 100000000
+max-tx-payload-size = 174763
+disabled-features = []
+
+[experiments]
```
Thoralf-M: Are these values needed or can we just delete them?

VmMad (approved):

dapp/tests/utils.ts
```diff
@@ -460,3 +481,26 @@ export async function createCoupon({ code, type, value }: CreateCouponOptions) {
         throw error;
     }
 }
+
+export async function checkAddressBalanceWithRetries(address: string): Promise<void> {
```
Thoralf-M, line 489: How about moving this into requestFaucetTokens()? So add an optional bool param and if true is provided, wait until there is balance

.github/scripts/setup-e2e-network.sh
msarcev: Those are defaults for the `config.toml` of the graphql process. Don't really care about them except that i need a valid config to pass the iota-names params along.
Possibly some values can be omitted but i didn't check. 😅

dapp/tests/utils.ts
msarcev: Sure

msarcev: https://github.com/iotaledger/iota-names/pull/1011/changes/22301356620608277e1119ec301be290157ad78e

.github/scripts/setup-e2e-network.sh
msarcev: @Thoralf-M You think that if we specify **only** the iota-names params in the config the code will take those and use the defaults for the rest ?
In reading the logic it didn't look like that.

Thoralf-M: I haven't checked how it works, but yeah I asked because it might be that it would just use this as default, if not it's fine to keep it as is

Thoralf-M (approved):

PR #1019, opened 2025-12-15
feat(dapp): name click on amplitude

### LINKED ISSUE #980: Name clicks in amplitude
We should name all clicks in the app so they are easier to categorize, see https://iotafoundation.atlassian.net/wiki/spaces/IOTANames/pages/2621931526/IOTA+Names+-+Amplitude+Events (under Second round section)

panteleymonchuk:
Closes #980


evavirseda (approved):

evavirseda (changes requested):
~~ci is failing 🙈~~

evavirseda (approved):

evavirseda (commented):
I’m checking the Amplitude dashboard because I’m thinking we might be able to rename the default events there 💡 we can discuss it on slack if you want 🙏

evavirseda (approved):
looks good!

dapp/src/lib/utils/analytics/amplitude.ts
```diff
@@ -9,10 +9,9 @@ import { CONFIG } from '@/config';
 
 import { ampli } from './ampli';
 import { AMP_COOKIES_KEY } from './constants';
+import { contextEnrichmentPlugin } from './plugins/contextEnrichmentPlugin';
 
-const IS_ENABLED =
-    process.env.NEXT_PUBLIC_BUILD_ENV === 'production' &&
-    process.env.NEXT_PUBLIC_AMPLITUDE_ENABLED === 'true';
+const IS_ENABLED = true;
```
evavirseda: looks like you forgot to revert it? 🤭 
```suggestion
const IS_ENABLED =
    process.env.NEXT_PUBLIC_BUILD_ENV === 'production' &&
    process.env.NEXT_PUBLIC_AMPLITUDE_ENABLED === 'true';
```

evavirseda (changes requested):

evavirseda (approved):
looks great 🙌 
snapshot is a nice fix for the modals, and i think using an allowlist is way cleaner than manually deleting keys

Juligs (approved):

PR #1041, opened 2026-01-08
feat(dapp): add to recent searches when performing automatic search

### LINKED ISSUE #1040: Improve recent searches history in the Availability Dialog
There's a problem right now: when performing a search without hitting the enter button the search is not added to the `Recent` pills, we should fix this

- Add it to the list after the debounced value exists and a search is done

VmMad:
Fixes #1040


KeitoTadashi (approved):
🎨

Juligs (approved):
:smile_cat:

PR #1056, opened 2026-01-12
chore(ci): Update nightly build version

DaughterOfMars:
## Description

Updates the nightly build version from `nightly-2025-04-01` -> `nightly-2026-01-07` in the CI.


Thoralf-M (approved):

indexer/deny.toml
```diff
@@ -117,6 +119,7 @@ allow = [
   "Zlib",
   "OpenSSL",
   "Apache-2.0 WITH LLVM-exception",
+  "CDDL-1.0",
```
thibault-martinez, line 122: In general we should be careful about adding stuff here. `CDDL-1.0` is not compatible with `Apache 2.0`. It's fine through dependencies but allowing it could make it so we're not warned about for legit incompatibility cases.

thibault-martinez (approved):

PR #1075, opened 2026-01-14
fix(dapp): Disable uppercase typing in auctions search

### LINKED ISSUE #1073: User reported upper case bug in auction search
## Bug Description

A user reported a bug in the auction search on iPhone where the second letter of a search term would be upper case.

## Steps To Reproduce

I don't seem to be able to reproduce exactly what the reported, but sometimes it happens that the first two chars are upper case.
What I can reproduce tho is:

1) Go to https://iotanames.com/auctions
2) Type on search
3) Disable upper case for the first later
4) Type two letters. The second one is upper case

## Expected Behaviour

Upper case should not be enabled after the first char.

marc2332:
Closes https://github.com/iotaledger/iota-names/issues/1073


KeitoTadashi (approved):

Dr-Electron (approved):

Dr-Electron (approved):
Still works :trollface:

PR #1077, opened 2026-01-14
feat(dapp): improve gas management on transactions

VmMad:
When buying a name that costs 50 IOTA and you only have 50 IOTA you will get an error that you cannot pay for the transaction but the cost remains 50 IOTA, which is unclear because we should include the gas price even if the user cannot pay for it


KeitoTadashi (approved):

dapp/src/components/dialogs/RenewNameDialog.tsx
```diff
@@ -232,6 +232,7 @@ export function RenewNameDialog({ setOpen, name, onRenew }: RenewDialogProps) {
     const disableEdit = isSendingTransaction || isSigning || renewOptions.length === 0;
     const disableSave = isLoading || !canRenew || !renewYears || !updateNameTransaction;
```
KeitoTadashi:
```suggestion
    const disableSave = isLoading || !canRenew || !renewYears || txError;
```

KeitoTadashi (changes requested):
disable button here, I was able to hit renew and approve the TX, then `renew succesfully` pop, but TX [failed](https://explorer-staging-iota1.vercel.app/txblock/5f9DU7FTEFcyBwSXYpLsa3sUGiaq5pAJhfow8NRvgEpv?network=devnet)

https://github.com/user-attachments/assets/eb7419cc-29ba-4559-a867-d41bc16eb5be

KeitoTadashi (dismissed):

brancoder (changes requested):
requested a change until we decide how to display the price

VmMad:
Not planned.

PR #1086, opened 2026-01-20
fix(ts-sdk): add retries when fetching balance data

### LINKED ISSUE #1045: Fix flaky ts-sdk tests
See https://github.com/iotaledger/iota-names/actions/runs/20785504833/job/59693784275
This issue seems to pop from time to time

VmMad:
Fixes #1045

Balance is probably not propagated since the faucet is correctly requested, so we need to give some time

A run that would have failed without the fix: https://github.com/iotaledger/iota-names/actions/runs/21175631179/job/61008204575#step:11:303

### DIFF
--- a/sdk/test/toolbox.ts
+++ b/sdk/test/toolbox.ts
@@ -71,9 +71,18 @@ export async function setupIotaClient() {
         logger: (msg) => console.warn('Retrying requesting from faucet: ' + msg),
     });
 
-    const b = await client.getBalance({
-        owner: address,
-    });
+    const b = await retry(
+        () =>
+            client.getBalance({
+                owner: address,
+            }),
+        {
+            retries: 10,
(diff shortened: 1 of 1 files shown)


sdk/test/toolbox.ts
```diff
@@ -71,9 +71,18 @@ export async function setupIotaClient() {
...
-        owner: address,
-    });
+    const b = await retry(
+        () =>
+            client.getBalance({
+                owner: address,
+            }),
+        {
+            retries: 10,
+            delay: 1000,
```
Thoralf-M: Would rather do something like this, 1 second delay is pretty long
```suggestion
            retries: 100,
            delay: 50,
```

Thoralf-M (approved):

KeitoTadashi (approved):

PR #1092, opened 2026-01-21
feat(dapp): add error case when creating a subname that is reserved

### LINKED ISSUE #1090: Improve error message for reserved/blocked subname creation
Lets say “amazon” is reserved and I own “cool.iota”, I can’t create “amazon.cool.iota”
In the dapp we currently fail with a very generic error "Something went wrong, try again later". But we should match the error in the dryrun and provide a better error eg "This subname can't be created because `{name_here}` is reserved or blocked"

i think we could try matching the error with `assert_not_blocked_or_reserved` in the error from the dry ryn

VmMad:
Fixes #1090

### DIFF
--- a/dapp/src/lib/constants/errorMessages.constants.ts
+++ b/dapp/src/lib/constants/errorMessages.constants.ts
@@ -7,6 +7,7 @@ export const INSUFFICIENT_COIN_BALANCE_ID = 'InsufficientCoinBalance';
 export const INVALID_IOTA_NAME = 'Invalid IOTA names';
 export const INVALID_IOTA_OBJECT_ID = 'Invalid IOTA Object id';
 export const INVALID_U64_VALUE = 'Invalid u64 value';
+const BLOCKED_OR_RESERVED_FUNCTION_NAME = 'assert_not_blocked_or_reserved';
 
 export const NO_BALANCE_GENERIC_MESSAGE = 'Make sure you have enough funds and try again.';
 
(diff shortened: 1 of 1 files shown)


dapp/src/lib/constants/errorMessages.constants.ts
```diff
@@ -7,6 +7,7 @@ export const INSUFFICIENT_COIN_BALANCE_ID = 'InsufficientCoinBalance';
 export const INVALID_IOTA_NAME = 'Invalid IOTA names';
 export const INVALID_IOTA_OBJECT_ID = 'Invalid IOTA Object id';
 export const INVALID_U64_VALUE = 'Invalid u64 value';
+const BLOCKED_OR_RESERVED_FUNCTION_NAME = 'assert_not_blocked_or_reserved';
```
KeitoTadashi:
```suggestion
export const BLOCKED_OR_RESERVED_FUNCTION_NAME = 'assert_not_blocked_or_reserved';
```

should we either remove or add `export` to all const in order to unify

KeitoTadashi (commented):
just a nit

dapp/src/lib/constants/errorMessages.constants.ts
KeitoTadashi: probably remove them since we export the big one which contains all consts

KeitoTadashi (approved):

evavirseda (approved):

PR #1095, opened 2026-01-23
feat(dapp): fix missing address in success modal chip

### LINKED ISSUE #1094: Fix address chip is empty in success modal (missing address)
in the success modal (after completing the action), the address chip renders with no text, the icons (link + copy) are visible, but the address label is empty

Image

evavirseda:
fixes https://github.com/iotaledger/iota-names/issues/1094


msarcev (approved):

dapp/src/components/dialogs/ConnectToAddressDialog.tsx
brancoder, line 366: we should update copy address code to use target address. now clicking on copy, after setting an address different from the current account, doesn't use the target address but current address

Image

brancoder (changes requested):
LG just the copy code needs updating

brancoder (approved):
LGTM

PR #1098, opened 2026-01-26
fix(dapp): avoid adding ampli plugins when ampli is disabled

VmMad:


begonaalvarezd:
Closing PR since its more convenient managing the amplitude plugins from amplitude directly

PR #1121, opened 2026-02-04
fix: `RUSTSEC-2026-0007`

thibault-martinez:


Alex6323 (approved):

indexer/Cargo.lock
```diff
@@ -1708,7 +1708,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7ab67060fc6b8ef687992d439ca0fa36e7ed17e9a0b16b25b601e8757df720de"
 dependencies = [
  "data-encoding",
- "syn 2.0.114",
+ "syn 1.0.109",
```
Thoralf-M: Why this downgrade?

thibault-martinez: I am not sure, I just did `cargo update -p bytes`. I can update all deps if you prefer.

Thoralf-M: It's just weird because this is 3 years old and looking at the `bytes` crate dependencies nothing seem to have changed there

Thoralf-M (approved):

indexer/Cargo.lock
thibault-martinez: I removed the lock file, did a general cargo update and rebuilt it now

PR #1134, opened 2026-02-09
set data retention to 100 years

shufps:
Set data retention to 100 years, effectively disabling auto-pruning


thibault-martinez (approved):

Thoralf-M (approved):

Alex6323 (approved):
are we sure that's enough?

PR #1140, opened 2026-02-10
chore(dapp): ensure session replay does not run on devmode

### LINKED ISSUE #1138: Ensure session replay does not run on devmode
We have to prevent the session replay to trigger even on devmode when devs have the amplitude env enabled

evavirseda:
fixes https://github.com/iotaledger/iota-names/issues/1138

### DIFF
--- a/dapp/src/lib/utils/analytics/amplitude.ts
+++ b/dapp/src/lib/utils/analytics/amplitude.ts
@@ -10,8 +10,11 @@ import { ampli } from './ampli';
 import { AMP_COOKIES_KEY } from './constants';
 import { contextEnrichmentPlugin } from './plugins/contextEnrichmentPlugin';
 
+// Dev note: do not change, production-only gate (Session Replay must never run outside prod)
 const IS_PRODUCTION = process.env.NEXT_PUBLIC_BUILD_ENV === 'production';
 
+// Dev note: set IS_ENABLED=true to test Amplitude events locally
+// Session Replay is still blocked unless IS_PRODUCTION is true
 const IS_ENABLED = IS_PRODUCTION && process.env.NEXT_PUBLIC_AMPLITUDE_ENABLED === 'true';
 
 /**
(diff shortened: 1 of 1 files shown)


panteleymonchuk (approved):

dapp/src/lib/utils/analytics/amplitude.ts
```diff
@@ -67,19 +67,21 @@ export async function initAmplitude() {
...
-                    maskSelector: [
-                        '.amp-obfuscation', // any element with this class will be masked according to the defaultMaskLevel
-                        // specific selectors for the dropdown menu (comes from @iota/dapp-kit) to ensure it's fully masked
-                        '[data-radix-popper-content-wrapper]',
-                        '[class*="AccountDropdownMenu"]',
-                    ],
-                },
-            });
-            ampli.client.add(sessionReplayTracking);
+            if (IS_PRODUCTION) {
```
begonaalvarezd, line 73: `IS_ENABLED` is defined as  `IS_PRODUCTION && somethingelse` so i dont see how this line changes anything

begonaalvarezd: The current codebase already ensures devs dont run this unless they build and run the app without devmode and also have the enabled env to true. So the code is already quite robust without the need of this PR. If anything, we could just check if the dapp is not running in `https://iotanames.com/*` and then maybe add `amplitude.remove('sessionReplayTracking')` if its not, but idk is just an idea

We can also just close this PR and the issue, i am just fearing devs running the app no tin devmode with the env enabled and then polluting all the session replays

begonaalvarezd (changes requested):

dapp/src/lib/utils/analytics/amplitude.ts
evavirseda: we set `IS_ENABLED` to true so we can test in amplitude, so that condition works like a double gate: if `IS_ENABLED` is turned on it enters the if, but since `IS_PRODUCTION` isn’t true, session replay won’t run

evavirseda: maybe I can add a comment for devs, like: just set `IS_ENABLED` to true for testing amplitude

begonaalvarezd: I see, lets add comments about this then

begonaalvarezd (approved):

PR #1151, opened 2026-02-19
feat(grafana): add purchases metric

Alex6323:
Adds a new metric (Number of New Direct IOTA Name purchases) to the Grafana dashboard.

Tested it by running the indexer against a localnet and monitored the grafana dashboard while directly purchasing names (counter increases) and subnames (counter doesn't increase).

### DIFF
--- a/indexer/src/metrics.rs
+++ b/indexer/src/metrics.rs
@@ -4,6 +4,8 @@
 use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 
 use axum::{Extension, Router, routing::get};
+use futures::StreamExt;
+use iota_graphql_client::{Client as GraphqlClient, query_types::EventFilter};
 use iota_metrics::{METRICS_ROUTE, RegistryService, histogram::Histogram};
 use prometheus::{
     IntCounter, IntCounterVec, IntGauge, IntGaugeVec, Registry,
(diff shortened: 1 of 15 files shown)


indexer/src/metrics.rs
```diff
@@ -208,6 +218,50 @@ impl IotaNamesMetrics {
         Ok(())
     }
 
+    /// Backfill the `new_purchases` counter by querying historical
+    /// `TransactionEvent` events from the IOTA GraphQL API and counting
+    /// those where `is_renewal` is false.
+    ///
+    /// This is a no-op if the counter already has a value (e.g. restored from
+    /// Prometheus).
+    pub async fn backfill_purchases_from_graphql(
```
thibault-martinez: Tbh I'm not sure we should have backfilling code, we'll just ask devops to resync

Alex6323: isn't that a huge pain just for this small metric?

Alex6323: backfilling takes <1s

thibault-martinez: > isn't that a huge pain just for this small metric?

Not really no

> backfilling takes <1s

It's more about having/keeping code that's not really needed

Alex6323: okay :thumbsup:

Alex6323: https://github.com/iotaledger/iota-names/pull/1151/commits/072608ecd2c2ee28b8bb38915f603aed3e85d202

Thoralf-M (approved):

indexer/src/metrics.rs
```diff
@@ -33,6 +33,7 @@ pub(crate) struct IotaNamesMetrics {
     pub auction_bid_count_distribution: IntGaugeVec,
     pub percentage_discounts: IntGauge,
     pub fixed_discounts: IntGauge,
+    pub new_purchases: IntCounter,
```
thibault-martinez: I would personally drop the `new_` everything is new

DaughterOfMars: Yeah probably not needed

DaughterOfMars (approved):

indexer/src/metrics.rs
Alex6323: changed 9f4a4615fa12dfaabdf466aff9898b507d6bfb1c

thibault-martinez (approved):

PR #1157, opened 2026-02-24
chore(dapp): Remove isConnected guard check in PurchaseName dialog

marc2332:
Useless, the dialog won't get open if the user is not connected, hence why we dont have this guard in the rest of dialogs


KeitoTadashi (approved):
🧹

msarcev (approved):

PR #1187, opened 2026-03-13
feat(dapp): add custom date picker for subnames

### LINKED ISSUE #915: Add custom date picker to subname renewal
Subnames can be renewed and any date can be set as expiration as soon as is less or equal than the father. We should allow selecting any date then

Additionally, we can add to the amplitude event a new property to specify if the expiration date is the same as the parent or custom, just to distinguish between those 2, lets not add the date itself

Depends on [design](https://github.com/iotaledger/if-comms/issues/2155)

KeitoTadashi:
fixes #915 

Question: `CreateSubnameDialog` L:186-192: do we want to add `expirationDate` as parameter in amplitude?
Response: We are going to add `expirationType: isParentExpiration ? 'parent' : 'custom',`
```ts
ampli.createdSubname({
       name: fullSubnameName,
       subname: editSubname,
       parentName: name,
    ➡️ expirationType: isParentExpiration ? 'parent' : 'custom', ⬅️ 
       allowToRenewExpiration: editIsAllowingRenew,
       allowToCreateAdditionalSubnames: editIsAllowSubnames,
});
```

### DIFF
--- a/dapp/src/components/ExpirationDate.tsx
+++ b/dapp/src/components/ExpirationDate.tsx
@@ -97,9 +97,6 @@ export function ExpirationDate({
                         ? new Date(minDate.getTime() + minimumDuration)
                         : new Date(now.getTime() + 2 * minimumDuration);
 
-                console.log('Selected Date:', selectedDate);
-                console.log('now:', now);
-                console.log('Minimum Date:', minimumDate);
                 if (selectedDate > maxDate) {
                     setError("Must be less than or equal to the parent name's date");
                     setCustomExpirationDate(null);
--- a/dapp/src/components/ExpirationDate.tsx
+++ b/dapp/src/components/ExpirationDate.tsx
@@ -12,9 +12,9 @@ import {
     Select,
     SelectOption,
 } from '@iota/apps-ui-kit';
-import { MINIMUM_SUBNAME_DURATION } from '@iota/iota-names-sdk';
 import { useCallback, useEffect, useMemo, useState } from 'react';
 
+import { useNamesConfig } from '@/hooks/useNamesConfig';
 import { formatExpirationDate } from '@/lib/utils/format/formatExpirationDate';
 
 const MONTH_NAMES = [
@@ -36,7 +36,7 @@ interface ExpirationDateProps {
     parentExpirationDate: Date | null;
     currentExpirationDate: Date | null;
     maxDate: Date | null;
-    minDate: Date | null;
+    minDate?: Date | null;
     onChange: (date: Date | null) => void;
     onExpirationTypeChange?: (isParentExpiration: boolean) => void;
 }
@@ -51,6 +51,8 @@ export function ExpirationDate({
     onChange,
     onExpirationTypeChange,
 }: ExpirationDateProps) {
+    const { data: config, isLoading: isLoadingConfig } = useNamesConfig();
+
     const [month, setMonth] = useState<number | null>(null);
     const [day, setDay] = useState<number | null>(null);
     const [year, setYear] = useState<number | null>(null);
@@ -69,31 +71,47 @@ export function ExpirationDate({
 
     const emitChange = useCallback(
         (mon: number | null, day: number | null, year: number | null) => {
-            if (mon !== null && day !== null && year !== null && maxDate !== null) {
-                const date = new Date(Date.UTC(year, mon, day));
+            if (
+                mon !== null &&
+                day !== null &&
+                year !== null &&
+                maxDate !== null &&
+                minDate !== null &&
+                !isLoadingConfig &&
+                config?.subnamesConfig
+            ) {
+                const minimumDuration = Number(config.subnamesConfig.minimum_duration);
                 const now = new Date();
-                const todayOnly = new Date(
-                    Date.UTC(now.getUTCFullYear(), now.getUTCMonth(), now.getUTCDate()),
+                const selectedDate = new Date(
+                    year,
+                    mon,
+                    day,
+                    now.getHours(),
+                    now.getMinutes(),
+                    now.getSeconds(),
+                    now.getMilliseconds(),
                 );
-                const minimumDate = minDate
-                    ? new Date(
-                          Math.max(minDate.getTime(), todayOnly.getTime()) +
-                              MINIMUM_SUBNAME_DURATION,
-                      )
-                    : todayOnly;
 
-                if (date > maxDate) {
+                const minimumDate =
+                    minDate && minDate > selectedDate
+                        ? new Date(minDate.getTime() + minimumDuration)
+                        : new Date(now.getTime() + 2 * minimumDuration);
+
+                console.log('Selected Date:', selectedDate);
+                console.log('now:', now);
+                console.log('Minimum Date:', minimumDate);
+                if (selectedDate > maxDate) {
                     setError("Must be less than or equal to the parent name's date");
                     setCustomExpirationDate(null);
                     onChange(null);
-                } else if (date < minimumDate) {
+                } else if (selectedDate < minimumDate) {
                     setError(`Must be ${formatExpirationDate(minimumDate)} or later`);
                     setCustomExpirationDate(null);
                     onChange(null);
                 } else {
                     setError(null);
-                    setCustomExpirationDate(date);
-                    onChange(date);
+                    setCustomExpirationDate(selectedDate);
+                    onChange(selectedDate);
                 }
             } else {
                 setError(null);
@@ -101,7 +119,7 @@ export function ExpirationDate({
                 onChange(null);
             }
         },
-        [onChange, maxDate],
+        [onChange, maxDate, minDate, config, isLoadingConfig],
     );
 
     const monthOptions: SelectOption[] = useMemo(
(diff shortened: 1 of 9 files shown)


dapp/src/components/dialogs/CreateSubnameDialog.tsx
```diff
@@ -239,6 +250,17 @@ export function CreateSubnameDialog({ name, setOpen }: CreateSubnameProps) {
                                           : undefined
                                 }
                             />
+                            <div className="flex flex-col gap-y-md w-full">
+                                <ExpirationDate
+                                    onChange={setExpirationDate}
+                                    parentExpirationDate={parentExpirationDate}
+                                    currentExpirationDate={null}
+                                    maxDate={
+                                        parentExpirationDate ? parentExpirationDate : new Date()
```
panteleymonchuk: in my understanding, `parentExpirationDate` should always exists, as we can't create subname without parent name.

Also, why we provide here `new Date()`?
`maxDate: Date | null`, `parentExpirationDate: Date | null` 
So why just don't provide `maxDate={parentExpirationDate}`?

dapp/src/components/ExpirationDate.tsx
```diff
@@ -0,0 +1,277 @@
...
+                    now.getHours(),
+                    now.getMinutes(),
+                    now.getSeconds(),
+                    now.getMilliseconds(),
+                );
+
+                const minimumDate =
+                    minDate && minDate > selectedDate
+                        ? new Date(minDate.getTime() + minimumDuration)
+                        : new Date(now.getTime() + 2 * minimumDuration);
```
panteleymonchuk, line 98: Could you remind what is this number about?

dapp/src/components/ExpirationDate.tsx
```diff
@@ -0,0 +1,277 @@
...
+                            disabled={isParentExpiration}
+                        />
+                    </div>
+                    <div className="flex-1">
+                        <Select
+                            options={dayOptions}
+                            value={day !== null ? String(day) : PLACEHOLDER_ID}
+                            onValueChange={handleChange('day')}
+                            disabled={isParentExpiration}
+                        />
```
panteleymonchuk, line 253: Can we make select scrollable?

https://github.com/user-attachments/assets/a1f02193-c69a-48da-b44c-af65f8e02661

panteleymonchuk (changes requested):

dapp/src/components/ExpirationDate.tsx
KeitoTadashi: That's the minimum time you can renew a subname. 
-> if you are creating, the minimum is 24h from the moment of renew
-> if you are renewing, the minimum is the previous date + 24 hours

When creating a new subname i'm setting 48 hours because there are some problems with hours since we set a time to renew, but we don't know when user is signing tx and chain clock is still ticking.

Simple example: 

1. it's 1 January 10:00 and user want to create a new subname with expiration 2 January. 
2. We set same time to the new subname so expiration time is 2 January at 10:00. 
3. User takes 1 minute to sign tx. 
4. Expiration time - Current time = 23:59 hours
4. On chain check it's going to fail because 24h > 23:59

dapp/src/components/ExpirationDate.tsx
KeitoTadashi: will need to revert my ''scrollable selects'' in favor of [this](https://github.com/iotaledger/iota/pull/10693)

msarcev:
Is it worth adding an e2e test that covers this feature ?

dapp/src/components/ExpirationDate.tsx
```diff
@@ -0,0 +1,277 @@
...
+    const setters: Record<DateField, (v: number) => void> = {
+        month: setMonth,
+        day: setDay,
+        year: setYear,
+    };
+
+    const handleChange = (field: DateField) => (value: string) => {
+        if (value === PLACEHOLDER_ID) return;
+        const num = Number(value);
+        setters[field](num);
```
msarcev, line 185: what is this `setters` call about ?

dapp/src/components/ExpirationDate.tsx
```diff
@@ -0,0 +1,277 @@
...
+                return {
+                    id: String(day),
+                    renderLabel: () => <span className="text-body-lg">{day}</span>,
+                };
+            }),
+        ],
+        [daysInMonth],
+    );
+
+    type DateField = 'month' | 'day' | 'year';
```
msarcev: Move this out please

dapp/src/components/ExpirationDate.tsx
msarcev: Nvm, i see.

dapp/src/components/ExpirationDate.tsx
```diff
@@ -0,0 +1,277 @@
...
+    const [isParentExpiration, setIsParentExpiration] = useState<boolean>(true);
+    const [customExpirationDate, setCustomExpirationDate] = useState<Date | null>(null);
+    const currentYear = new Date().getFullYear();
+
+    const parentTime = parentExpirationDate?.getTime() ?? null;
+    useEffect(() => {
+        if (isParentExpiration && parentExpirationDate) {
+            onChange(parentExpirationDate);
+        }
+    }, [parentTime, isParentExpiration]);
```
msarcev, line 71: Should we include `onChange` here ?

dapp/src/components/ExpirationDate.tsx
```diff
@@ -0,0 +1,277 @@
...
+                    mon,
+                    day,
+                    now.getHours(),
+                    now.getMinutes(),
+                    now.getSeconds(),
+                    now.getMilliseconds(),
+                );
+
+                const minimumDate =
+                    minDate && minDate > selectedDate && minDate > now
```
msarcev, line 96: Is `minDate > selectedDate` correct and needed here ?

minDate (current subname expiration) being greater than selected date means the selected date is invalid, right ?
Why would the renewal branch be used only if that is true ?

dapp/src/components/ExpirationDate.tsx
```diff
@@ -0,0 +1,277 @@
...
+    }, [parentTime, isParentExpiration]);
+
+    const emitChange = useCallback(
+        (mon: number | null, day: number | null, year: number | null) => {
+            if (
+                mon !== null &&
+                day !== null &&
+                year !== null &&
+                maxDate !== null &&
+                minDate !== null &&
```
msarcev: `minDate?: Date | null;` can also be `undefined` so maybe we should cover that case here.

msarcev (changes requested):

dapp/src/components/ExpirationDate.tsx
KeitoTadashi: Yes, it is needed because I want to suggest to the user which date is available for name/subname

```ts
setError(`Must be ${formatExpirationDate(minimumDate)} or later`);
```

dapp/src/components/ExpirationDate.tsx
KeitoTadashi: yes, it is for initialize component, otherwise it came null and "same expiration as parent" pops

imagen

msarcev: I meant to include `onChange` in the dependency array.

VmMad (dismissed):
Image
it shouldn't be possible to select a day/month higher than the parent nft. unless it shows an error

dapp/src/components/ExpirationDate.tsx
KeitoTadashi: actually I removed the condition `minDate !== null` because it can never be null, and undefined is used in
```js
const minimumDate =
    minDate && minDate > selectedDate && minDate > now
        ? new Date(minDate.getTime() + minimumDuration)
        : new Date(now.getTime() + 2 * minimumDuration);
```

KeitoTadashi:
> Image it shouldn't be possible to select a day/month higher than the parent nft. unless it shows an error

@VmMad thanks, my last commit had a bad condition which I just reverted

dapp/src/components/ExpirationDate.tsx
```diff
@@ -0,0 +1,276 @@
...
+    }, [parentTime, isParentExpiration]);
+
+    const emitChange = useCallback(
+        (mon: number | null, day: number | null, year: number | null) => {
+            if (
+                mon !== null &&
+                day !== null &&
+                year !== null &&
+                maxDate !== null &&
+                !isLoadingConfig &&
```
msarcev, line 80: You didn't revert, you removed the `min != null` entirely. We don't need it then ?

KeitoTadashi: no, we don't need it anymore

dapp/src/components/ExpirationDate.tsx
KeitoTadashi: no, we should not do that because the effect only needs to re-run when the user toggles back to "Same as parent"

msarcev (approved):

evavirseda (approved):

panteleymonchuk (approved):

PR #1198, opened 2026-03-31
chore(names-sdk): Rework names sdk release workflow

marc2332:
Inspired by https://github.com/iotaledger/iota/pull/10906

### DIFF
--- a/.github/workflows/changesets_publish.yml
+++ b/.github/workflows/changesets_publish.yml
@@ -3,16 +3,16 @@ name: Changesets Publish
 on:
   workflow_dispatch:
     inputs:
-      tag:
-        description: "Release tag (e.g., latest, experimental, dev)"
+      release_type:
+        description: "Release type"
         required: true
-        type: string
-      create_snapshot:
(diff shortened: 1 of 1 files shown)


msarcev (approved):

.github/workflows/changesets_publish.yml
```diff
@@ -48,46 +50,59 @@ jobs:
           fi
 
       - name: Install dependencies
+        if: github.event.inputs.release_type != 'latest'
         run: pnpm install --frozen-lockfile
 
+      - name: Create snapshot versions
+        if: github.event.inputs.release_type == 'experimental'
+        run: pnpm changeset version --snapshot experimental
```
Thoralf-M, line 58: How are version set for `next` instead of `experimental`?

msarcev: We do it with a PR that bumps the version in package.json based on changeset files.
So we bump manually explicitly.

`experimental` is meant to be run on a branch that has changeset files on it, so that is does the versioning with `experimental-x.y.z...`

Thoralf-M (approved):

PR #1201, opened 2026-04-02
chore(ci): Unify iota binary download in ts ci

marc2332:
Doesnt make sense that we use homebrew for sdk tests but for dapps we download.


.github/workflows/typescript.yml
Thoralf-M, line 118: iota-localnet binary is needed for this now (from v1.20)

marc2332: no, wrong, we are still using 1.19

marc2332: I will change to use the iota-localnet binary soon though

Thoralf-M (approved):

thibault-martinez (approved):

